PR #16, opened 2013-07-04
FilesystemCache should not use that many directories

nesQuick:
... in order to have a manageable amount of items per directory. [See this blog post.](http://michaelandrews.typepad.com/the_technical_times/2009/10/creating-a-hashed-directory-structure.html)

When splitting the hash into four pieces, each containing 16 chars, with each 16 possible values this can lead to 16^16 (1.84467441e19) folders at each level. This might lead to an amount which goes far beyond filesystem limits. Even if a filesystem could handle that many items/folders, "stat operations such as listing files, following paths, or checking for the existence of a file" will be super slow!

### DIFF
--- a/lib/Doctrine/Common/Cache/FileCache.php
+++ b/lib/Doctrine/Common/Cache/FileCache.php
@@ -97,7 +97,7 @@ abstract class FileCache extends CacheProvider
     protected function getFilename($id)
     {
         $hash = hash('sha256', $id);
-        $path = implode(str_split($hash, 16), DIRECTORY_SEPARATOR);
+        $path = implode(str_split($hash, 2), DIRECTORY_SEPARATOR);
         $path = $this->directory . DIRECTORY_SEPARATOR . $path;
         $id   = preg_replace('@[\\\/:"*?<>|]+@', '', $id);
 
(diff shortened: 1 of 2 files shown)


Ocramius:
Is this still valid?
If so, I will gladly rebase and merge.

nesQuick:
Guess so under the constraint that ext2/ext3 did not change :)

Ocramius:
@nesQuick I was more wondering about any code change to be applied because of changed architecture/etc ;-)

nesQuick:
alright, no this change should fit as it's very small. Basically it's just changing a number ;D

Ocramius:
Merged into `master` at doctrine/cache@a335bf993de35ac24820ae25062d29a6e1d23877.

Thanks @nesQuick!

nesQuick:
:+1:

till:
:bow:

djoos:
First of all: great PR!

Unfortunately, I'm not really sure why this went out as 1.4: even though it doesn't break backward compatibility from a coding point of view, it does break backward compatibility from a cache point of view...

Anyway, we've updated our composer.json's doctrine/cache from 1.* to 1.3.* for now :-)

Ocramius:
@djoos you're supposed to nuke your cache at every upgrade of anything anyway.

djoos:
@Ocramius, I don't fully agree with that. What if you were to be caching nonces (WSSE)? You don't want to delete these everytime when rolling out a new version of your app and leave yourself vulnerable to replay attacks...

Ocramius:
@djoos you actually do kill your cache anyway, since even just our checksum algo could contain the cache library version in it.

Otherwise, simply do not upgrade.

PR #24, opened 2013-09-27
Cache namespace component

FabioBatSilva:
It requires a BC break..

The namespace is no logger added by default, 
it means that by default entries are stored without any prefix.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,8 +12,3 @@ Cache component extracted from the Doctrine Common project.
 
 * Added support for MongoDB as Cache Provider
 * Fix namespace version reset
-
-### v1.3
-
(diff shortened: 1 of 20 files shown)


Ocramius:
@FabioBatSilva can this be rebased? Is it actual?

ivoba:
+1 
This would make it possible to get the NamspacedKey from outside.

    $cache->getCacheNamespace()->getNamespacedKey($id)

I just have the case where i need this to bypass the cache to increase EXPIRE when using redis as backend.

Ocramius:
Closing: can be continued (if interesting) for 2.x

PR #30, opened 2014-03-26
Added support for setting the default file and folder permissions

kamermans:
After lecturing my developers on not using `0777` file permissions, and then forcing them to use the `Doctrine\Cache` component (because I love it :) ), I have come to realize that it creates `0777` folders, and this behavior is not adjustable.  The file permissions are determined by the current process umask as they are not explicitly set.  

This PR allows you to set the default folder and file mode if you so choose.  By default the existing behavior is used.

In my case, I would prefer `0770` for directories and `0660` for files.  Here's how to do it with my PR:

```php
$cache = new FilesystemCache('/tmp/');
$cache->setDirectoryMode(0770);
$cache->setFileMode(0660);
```

The specific use case for me is that my developers are in the `www-data` group, which is the group that creates the cache files, and I want them to be able to delete the cache files if need be, but I do not want the cache files to be readable by any random user on the system.

I have also added a unit test to cover the file permissions.  This test is skipped on Windows, since file permissions are not actually set (although `chmod` returns true and `fileperms` always returns `0644`).

### DIFF
--- a/lib/Doctrine/Common/Cache/FileCache.php
+++ b/lib/Doctrine/Common/Cache/FileCache.php
@@ -46,7 +46,7 @@ abstract class FileCache extends CacheProvider
      *
      * @var int
      */
-    protected $directory_mode = 0777;
+    protected $directoryMode = 0777;
 
     /**
      * The mode that files will be created with.  Null means the file will be created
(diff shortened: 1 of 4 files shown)


Ocramius:
@kamermans this PR is a bit messed up due to the fact that it's opened from your `master` branch.

Please re-open it from a feature branch.

kamermans:
Sorry about the branch mix-up, I'm pulling it up to the current master and making a new PR on a feature branch.

Ocramius:
@kamermans thanks for your understanding.

PR #29, opened 2014-02-14
Added multi get Interface and some basic implementations

goetas:
Added a multi-get interface that allows to benefit form vendors that offers a way to retrieve multiple items in one call (redis, memcached, apc...). 

I also added a default implementation into `CacheProvider` that fallbacks on `fetch()`.
I also added a multi-get implementation for memcache and apc.

PR #34, opened 2014-04-30
Added CacheMultiple interface

dragonwize:
Adding a CacheMultiple interface to ensure we can depend on multiples methods as suggested in #33.

See #33
See doctrine/DoctrineCacheBundle#16


Ocramius:
I handled fetches in #29: I suggest re-writing this PR to segregate write operations into single interfaces for now.

Ocramius:
Closing, as I already merged #29 and this would need adaptation anyway

PR #35, opened 2014-05-12
ElasticSearch Doctrine Cache Adapter

eddiejaoude:


Ocramius:
I'll rebase on merge.

Ocramius:
Decided not to merge this, as it introduces a soft-dependency to another external library.

I suggest moving it to a bridge library not maintained by us, as we can't ensure stability of this component.

Ocramius:
@eddiejaoude note that the code is perfectly ok, I just suggest to move this branch out of `doctrine/cache` and into a new library that depends (strictly) on `doctrine/cache` and the elasticsearch PHP client.

eddiejaoude:
@Ocramius sure no problem.

PR #48, opened 2014-10-04
Interface normalization.

Adirelle:
Although the CacheProvider provides several interesting capabilities, most are not available through the Cache interface. This PR simply declares new interfaces and CacheProvider as implementation of theses.

Note that FlushableCache could be implemented only by implementations that were actually able to flush all entries, instead of CacheProvider.

### DIFF
--- a/lib/Doctrine/Common/Cache/NamespacedCache.php
+++ b/lib/Doctrine/Common/Cache/NamespacedCache.php
@@ -24,12 +24,7 @@ namespace Doctrine\Common\Cache;
  *
  * @link   www.doctrine-project.org
  * @since  2.0
- * @author Benjamin Eberlei <kontakt@beberlei.de>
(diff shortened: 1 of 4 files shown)


lib/Doctrine/Common/Cache/NamespacedCache.php
```diff
@@ -0,0 +1,46 @@
...
+namespace Doctrine\Common\Cache;
+
+/**
+ * Interface for caches with namespace.
+ *
+ * @link   www.doctrine-project.org
+ * @since  1.4
+ * @author Adirelle <adirelle@gmail.com>
+ */
+interface NamespacedCache extends Cache
```
Ocramius: This usage was removed in doctrine/doctrine2#1259: I will remove the interface on merge.

Ocramius:
@Adirelle merged, thanks!

PR #50, opened 2014-11-04
FilesystemCache sanitizes filenames before writing

joe-walker:
Using the provided filename as the prefix for tempnam() lead to not being able to write the cache file if the filename had illegal characters. Characters other than letters, numbers, dashes, underscores, and full stops are now replaced with a dash.

### DIFF
--- a/lib/Doctrine/Common/Cache/FilesystemCache.php
+++ b/lib/Doctrine/Common/Cache/FilesystemCache.php
@@ -111,8 +111,10 @@ class FilesystemCache extends FileCache
         } elseif ( ! is_writable($filepath)) {
             return false;
         }
+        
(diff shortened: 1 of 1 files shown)


lib/Doctrine/Common/Cache/FilesystemCache.php
```diff
@@ -111,8 +111,10 @@ protected function doSave($id, $data, $lifeTime = 0)
         } elseif ( ! is_writable($filepath)) {
             return false;
         }
+        
+        $sanitizedFilename = preg_replace('/[^a-zA-Z0-9-_\.]/','-', basename($filename));
```
Ocramius: Merging this: working on a test case.

Ocramius:
I wrote more extensive tests and merged+rebased this locally. See doctrine/cache@907393ce5a36c1c4b4d0e6e8395fcd71fe6abd6c

Thanks @joe-walker!

PR #55, opened 2014-12-18
Fix race condition in PhpFileCache

tsteur:
* Applied fix for possible race conditions from doctrine/cache#38 to `PhpFileCache` by extracting the code into the parent class to make sure both concrete caches create the cache file the same way
* Added a line to `unlink` a created temp file in case `rename` was not successful
* Removed duplicated code and made it more consistent. For example the `FilesystemCache` tried to create a missing path in `doSave` whereas `PhpFileCache` did not.

If you prefer two or three separate pull requests for this let me know.

Issue was originally reported in piwik/piwik#6873

### DIFF
--- a/lib/Doctrine/Common/Cache/FileCache.php
+++ b/lib/Doctrine/Common/Cache/FileCache.php
@@ -146,6 +146,58 @@ abstract class FileCache extends CacheProvider
     }
 
     /**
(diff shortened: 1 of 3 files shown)


Ocramius:
@tsteur thanks, merged!

PR #56, opened 2015-01-11
Fixed some PSR2 violations

marcelsud:


lib/Doctrine/Common/Cache/FileCache.php
```diff
@@ -51,14 +51,14 @@
      */
     public function __construct($directory, $extension = null)
     {
-        if ( ! is_dir($directory) && ! @mkdir($directory, 0777, true)) {
+        if (! is_dir($directory) && ! @mkdir($directory, 0777, true)) {
```
Ocramius: We actually use the `if ( ! $foo` syntax in doctrine, that's a desired deviation from PSR-2

lib/Doctrine/Common/Cache/XcacheCache.php
```diff
@@ -86,7 +86,10 @@ protected function doFlush()
     protected function checkAuthorization()
     {
         if (ini_get('xcache.admin.enable_auth')) {
-            throw new \BadMethodCallException('To use all features of \Doctrine\Common\Cache\XcacheCache, you must set "xcache.admin.enable_auth" to "Off" in your php.ini.');
+            throw new \BadMethodCallException(
```
marcelsud: @Ocramius I see. Is this one also a desired deviantion?

Ocramius: This change is fine

PR #58, opened 2015-01-15
Improved support for Redis serializing against HHVM

o:
Removed support for Redis::SERIALIZER_IGBINARY for HHVM compability.

This fixes failing tests and functionality against HVVM with current master and previous versions. Tested with HHVM 3.3.0, Debian 7.6 and a running app top of Symfony 2.5.

Current cache driver implementation not works with HHVM, due to serialization problems:
https://travis-ci.org/doctrine/cache/jobs/46654706#L245-L358

Additional links for HHVM changes:
https://reviews.facebook.net/D31353
https://github.com/facebook/hhvm/pull/4623 https://github.com/facebook/hhvm/commit/c7cb8899a65b68984ca72aaa9fdbeb8236343535

### DIFF
--- a/lib/Doctrine/Common/Cache/RedisCache.php
+++ b/lib/Doctrine/Common/Cache/RedisCache.php
@@ -126,6 +126,9 @@ class RedisCache extends CacheProvider
      */
     protected function getSerializerValue()
     {
+        if (defined('HHVM_VERSION')) {
+            return Redis::SERIALIZER_PHP;
+        }
         return defined('Redis::SERIALIZER_IGBINARY') ? Redis::SERIALIZER_IGBINARY : Redis::SERIALIZER_PHP;
     }
 }


stof:
well, I think using Redis::SERIALIZER_IGBINARY when possible is still a good idea. If HHVM advocated support for IGBINARY while not supporting it, the logic choosing the serialization mode should be updated instead of removing it entirely.

o:
@stof Here is the alternative version supports detecting HHVM version https://gist.github.com/o/27a7e97c8082025bfd4b#file-gistfile1-php-L127-L134 

Looks like, it works.

stof:
please update your PR

lib/Doctrine/Common/Cache/RedisCache.php
```diff
@@ -126,6 +126,10 @@ protected function doGetStats()
      */
     protected function getSerializerValue()
     {
-        return defined('Redis::SERIALIZER_IGBINARY') ? Redis::SERIALIZER_IGBINARY : Redis::SERIALIZER_PHP;
+        if (defined('HHVM_VERSION')) {
+            return Redis::SERIALIZER_PHP;
+        } else {
```
stof: not need to use ``else`` as the if returns

o: Fixed

stof:
Looks like the Redis support in HHVM is stil broken

o:
I can't test with 3.4.0 (Installed version in Travis). 
But 3.3.0, 3.4.2 and 3.5.0-dev (rel) gd4d6687e3f34adcdf6dd89a57eab7e8bd08aa35b works without problem. 

My environment(s): Debian 7.6 (production), Ubuntu 14.04 (vagrant) and 14.10 (local), Symfony 2.5 (running app in production), using cache drivers through doctrine-cache-bundle.

Fi

Ocramius:
The travis version is indeed a bit out of date. Merging here after having checked locally against:

```
HipHop VM 3.5.0-dev (rel)
Compiler: heads/master-0-g7024026e00dec2fe6ca5ea46e23e5d5181648eef
Repo schema: 75a53eceff2c3914f29531368d9d363bf9507a35
Extension API: 20140829
```

o:
Thanks \o/

rybakit:
Actually, this fix is superfluous since the mentioned commit (https://github.com/facebook/hhvm/commit/c7cb8899a65b68984ca72aaa9fdbeb8236343535) has been merged into HHVM master.

Ocramius:
Master is still far from release: consider reverting this when HHVM 2.6 is
released.

Marco Pivetta

http://twitter.com/Ocramius

http://ocramius.github.com/

PR #60, opened 2015-01-16
Redis  MGET support

goetas:
Added redis and predis MGET support

### DIFF
--- a/lib/Doctrine/Common/Cache/PredisCache.php
+++ b/lib/Doctrine/Common/Cache/PredisCache.php
@@ -46,8 +46,7 @@ class PredisCache extends CacheProvider
     {
         $fetchedItems = call_user_func_array(array($this->client, 'mget'), $keys);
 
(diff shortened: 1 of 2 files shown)


guilhermeblanco:
Patch is ok, but I have something needling my head this seems wrong.
What's the idea here? If we fetch N keys and cache only returns < N, you need to fill out the blanks with null values?

goetas:
Looking at [this](https://github.com/doctrine/cache/blob/92642fc4902f1d4ac1b0e645152e833816767351/lib/Doctrine/Common/Cache/MultiGetCache.php#L36) the Predis implementation is correct, while the Redis implementation is not correct... i will fix it.

guilhermeblanco:
array_filter should do the work then... =)

goetas:
done

Ocramius:
Very good: merging, but it will be in `1.5.0` only.

PR #62, opened 2015-01-18
Bugfix: data being truncated in sqlite3 [WIP]

kelvinj:
- [ ] write unit test case to prove issue
- [x] write fix - bind data as SQLITE3_BLOB


Ocramius:
@kelvinj `str_repeat('a', 10000)` in a test? ;-)

Ocramius:
Cannot reproduce with testing (both with null bytes and with long strings, 10M+): closing as not valid, since this is most probably related with your library internal version (upgrade)

PR #63, opened 2015-01-19
Feature/mode spread hash setters

kamermans:
Here is the new PR based on https://github.com/doctrine/cache/pull/30 including configurable hashing mechanism, number of directory spread characters and file permissions.

Note that using SHA1 is causing one of my busy applications to fail in production due to a lack of inodes.  This is because there are so many directories created.  We are using this in production to overcome the issue:

```
$cache->setHasher('md5');
$cache->setDirectorySpreadChars(4);
```

This way we can balance the number of subdirs (4 deep max), with the number of files in each dir (10000 max, although much less in practice).

### DIFF
--- a/lib/Doctrine/Common/Cache/FileCache.php
+++ b/lib/Doctrine/Common/Cache/FileCache.php
@@ -46,7 +46,7 @@ abstract class FileCache extends CacheProvider
      *
      * @var int
      */
-    protected $directory_mode = 0777;
+    protected $directoryMode = 0777;
 
     /**
      * The mode that files will be created with.  Null means the file will be created
@@ -54,7 +54,7 @@ abstract class FileCache extends CacheProvider
      *
      * @var int|null
      */
-    protected $file_mode;
+    protected $fileMode;
 
     /**
      * Constructor.
@@ -66,7 +66,7 @@ abstract class FileCache extends CacheProvider
      */
     public function __construct($directory, $extension = null)
     {
-        if ( ! is_dir($directory) && ! @mkdir($directory, $this->directory_mode, true)) {
+        if ( ! is_dir($directory) && ! @mkdir($directory, $this->directoryMode, true)) {
             throw new \InvalidArgumentException(sprintf(
                 'The directory "%s" does not exist and could not be created.',
                 $directory
@@ -90,7 +90,10 @@ abstract class FileCache extends CacheProvider
      */
     public function setDirectoryMode($mode)
     {
-        $this->directory_mode = $mode;
+        if ( ! is_int($mode)) {
+            throw new \InvalidArgumentException("You must specify permissions modes as an int");
+        }
+        $this->directoryMode = $mode;
     }
 
     /**
@@ -99,7 +102,10 @@ abstract class FileCache extends CacheProvider
      */
     public function setFileMode($mode)
     {
-        $this->file_mode = $mode;
+        if ( ! is_int($mode)) {
+            throw new \InvalidArgumentException("You must specify permissions modes as an int");
+        }
+        $this->fileMode = $mode;
     }
 
     /**
--- a/lib/Doctrine/Common/Cache/FileCache.php
+++ b/lib/Doctrine/Common/Cache/FileCache.php
@@ -47,6 +47,15 @@ abstract class FileCache extends CacheProvider
      * @var int
      */
     protected $directoryMode = 0777;
+    
+    /**
+     * Cached objects are stored in directories.  These directory names are determined
+     * by splitting up the 32-char ID of the object into x parts.  If x=16 files will
+     * be two dirs deep (32/16=2).  If x=2 files will be eight dirs deep (32/2=16)
+     * 
+     * @var int
+     */
+    protected $directorySpreadChars = 16;
 
     /**
      * The mode that files will be created with.  Null means the file will be created
@@ -107,6 +116,22 @@ abstract class FileCache extends CacheProvider
         }
         $this->fileMode = $mode;
     }
+    
+    /**
+     * Cached objects are stored in directories.  These directory names are determined
+     * by splitting up the 32-char ID of the object into x parts.  
+     *   If x=16 files will be two dirs deep (32/16=2 ex: 1234567890123456/1234567890123456)
+     *   If x=2 files will be eight dirs deep (32/2=16 ex: 12/34/56/78/90/12/34/56/12/34/56/78/90/12/34/56)
+     * 
+     * @var int
+     */
+    public function setDirectorySpreadChars($chars)
+    {
+        if ( ! is_int($chars)) {
+            throw new \InvalidArgumentException("You must specify directory spread chars as an int");
+        }
+        $this->directorySpreadChars = $chars;
+    }
 
     /**
      * Gets the cache directory.
@@ -136,7 +161,7 @@ abstract class FileCache extends CacheProvider
     protected function getFilename($id)
     {
         $hash = hash('sha256', $id);
-        $path = implode(str_split($hash, 16), DIRECTORY_SEPARATOR);
+        $path = implode(str_split($hash, $this->directorySpreadChars), DIRECTORY_SEPARATOR);
         $path = $this->directory . DIRECTORY_SEPARATOR . $path;
         $id   = preg_replace('@[\\\/:"*?<>|]+@', '', $id);
 
--- a/lib/Doctrine/Common/Cache/FileCache.php
+++ b/lib/Doctrine/Common/Cache/FileCache.php
@@ -68,7 +68,7 @@ abstract class FileCache extends CacheProvider
      * 
      * @var int
      */
-    protected $directorySpreadChars = 16;
+    protected $directorySpreadChars = 2;
 
     /**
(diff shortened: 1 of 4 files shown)


Ocramius:
>  We are using this in production to overcome the issue

This raises a big question: why are you using a file-based cache on production?

This kind of cache is meant for shared hosting environments with very limited requirements: why are you not using an actual cache service instead?

That said, I'd rather write down the actual limitations of the `FileCache` than making it super-configurable.

kamermans:
I am using a file-based cache because I am saving image data.

lib/Doctrine/Common/Cache/FileCache.php
```diff
@@ -55,6 +55,37 @@
     private $replacementCharacters = array('__', '-');
 
     /**
+     * The mode that directories will be created with.
+     *
+     * @var int
+     */
+    protected $directoryMode = 0777;
```
stof: all these should be private

lib/Doctrine/Common/Cache/RedisCache.php
```diff
@@ -142,6 +142,7 @@ protected function doGetStats()
      */
     protected function getSerializerValue()
     {
+        // Only PHP Serialization is supported in HHVM
```
stof: the actual reason for this check is that only PHP serialization is supported, but HHVM 3.4 and older are still defining the constant, thus breaking the feature detection

lib/Doctrine/Common/Cache/FileCache.php
```diff
@@ -78,8 +109,60 @@ public function __construct($directory, $extension = '')
...
-        $this->extension = (string) $extension;
+        $this->extension = $extension ?: $this->extension;
+        $this->fileMode = 0666 & ~umask();
+    }
+
+    /**
+     * Sets the mode that new directories will be created with.
+     * @param int $mode Mode, normally in octal (ex: 0755)
+     */
+    public function setDirectoryMode($mode)
```
stof: do we need a setter for this ? the directory mode will end up being 0777 minus the umask because of the ``mkdir`` behavior (and if you configure it as 0775, it will still actually be 0755 minus the umask). I think it is easier to assume that you can set the umask to the value you need if you have special needs (this is what Symfony assumes for instance)

lib/Doctrine/Common/Cache/FileCache.php
```diff
@@ -55,6 +55,37 @@
...
+     */
+    protected $directorySpreadChars = 2;
+
+    /**
+     * The mode that files will be created with.  Null means the file will be created
+     * with the current umask.
+     *
+     * @var int|null
+     */
+    protected $fileMode;
```
stof: this file mode is not used currently, and the directory mode is not used everywhere. I suggest dropping them (see my comment on the ``setDirectoryMode`` method for more details about my reasons)

lib/Doctrine/Common/Cache/RedisCache.php
kamermans: Ok, so why is this check here then?  My severs are running HHVM 3.4.0 and 3.5.0 and neither of them seem to define HHVM_VERSION, so should this check be removed entirely and just `'return Redis::SERIALIZER_PHP;`?

lib/Doctrine/Common/Cache/FileCache.php
kamermans: I'll change them once we figure out what is going on with the rest of this PR.

lib/Doctrine/Common/Cache/FileCache.php
kamermans: My main issue is the umask is a convoluted mess so people will not use it, and in any case, the default file mode of 0777 is even worse.  At a minimum, perhaps the default dir perms should be 0775 and the default file perms 0664 so the cache files are not world-executable.  Are you suggesting we have a umask setter on this class, or that the user should change his umask before using this class?  Changing the global umask hardly seems like a good idea, especially in a multithreaded environment (`umask()` is not thread-safe).  This method would also make it impossible to have two cache instances with different permissions.  Setting the umask for this cache adapter itself (like my proposed mode setters) makes some sense, although I really don't see the point of having the user set the umask when they don't even know the base mode.  Furthermore, my issue with the cache files being executable remains unless you did something like `$dirUmask = 0022; $fileUmask = 0133;`.

kamermans:
@Ocramius on my 64GB partition I have 4194304 inodes (this is a factor of the filesystem and the inode-ratio, but mine are defaults for Ubuntu 14.04 with ext4).  With SHA1 hashing and 2-char spread (directories of 2 chars in length), you could create 131072 SHA1-based folders (out of the 4294967296 possible folders that SHA1-spread2 can produce) before you run out of inodes.  If there was 1 file in each folder tree, you would end up with a limit of 65536 cached items, although in practice 0.003% of the folders would be reused (that's 2 in my case), so you might squeeze 65538 items on a 64GB drive.

This ratio (64GB:64K items) is a reflection of the ext bytes-per-inode ratio (see `man mkfs.ext4 | grep -n3 bytes-per-inode`).

As another example, my backup drive is 2.7TB and has 183M inodes and so I should be able to store 1.4M items on it.  This ratio is different here because this drive has a higher bytes-per-inode ratio.

Note that using `md5` instead of `sha1` will almost double the max number of items that can be stored since the hash is half length.

Ocramius:
Makes sense, but I'm still wondering why you are using a file-based cache.

The fact is that the file cache was originally provided only as a "poor
man's cache solution" for those folks that are stuck on shared hosting, and
a lot of folks seem to use it and actually reach its limits.

That said, the PR seems legit, I'm just questioning all this fiddling
around the file-based caches :-)

Marco Pivetta

http://twitter.com/Ocramius

http://ocramius.github.com/

kamermans:
Using files is the only reasonable option, and I'm using Doctrine because it saves me the hasle of writing a new directory spreading algorithm and unit testing it, etc (granted that would not be terribly difficult).  It may have started as a poor man's caching solution, and I totally understand that with IO contention and such, however, I run multiple sites that get over 1B requests per month and they all use Doctrine for some sort of caching, and one uses the file cache, so I figured I would share the love and contribute back some of the improvements that we have found helpful.

Another argument for it is that it's nice to be able to drop in file cache on a system that would normally use something like Redis so you can limit the dependencies.  I've found myself jumping on a non-dev machine needing to test something before where it was nice to switch the cache type and be done with it :)

lib/Doctrine/Common/Cache/FileCache.php
```diff
@@ -78,8 +109,60 @@ public function __construct($directory, $extension = '')
...
+            throw new \InvalidArgumentException("You must specify permissions modes as an int");
+        }
+        $this->directoryMode = $mode;
+    }
+
+    /**
+     * Sets the mode that new files will be created with.
+     * @param int $mode Mode, normally in octal (ex: 0644)
+     */
+    public function setFileMode($mode)
```
Ocramius: Can you remove all these setters, and move them to the constructor instead?

lib/Doctrine/Common/Cache/RedisCache.php
Ocramius: This file is unrelated with the PR, we should leave it alone here.

lib/Doctrine/Common/Cache/FileCache.php
stof: @kamermans the default umask on all distributions I know of is ``0022``, which means that the directory will be created as ``0755`` unless you explicitly decide to change it. And even with your own config, the umask is still applied for ``mkdir`` calls.

And actually, for my local setup, I'm rather using ACLs to allow both PHP-FPM and my CLI to write to the same cache folders rather than changing permissions

stof: and cache files are not executable. Look at the current code. It uses ``0777`` as base for folders but ``0666`` for files

kamermans: Wow, I've totally missed the fact that the default umask is 022, it seems you're correct and that the mode setters are indeed unhelpful.

kamermans:
At this point I have asked my devs to just extend this project instead of contributing back as we seem to be spinning our wheels here.  I'm afraid my time is quite limited, so I'm not sure I can continue iterating here.  In any case, please consider that with the current system of sha1 with 2-char spread, any disk will run out of inodes *way* before it runs out of space.

Ocramius:
I can probably merge the hash algorithm change without the mask change...

Tobion:
I was really surprised that the FileCaches creates such a deep directory structure. According to twigphp/Twig#1684 this has negative impact. And twig only used 2 directories with 2 chars, but doctrine uses alot more!

jkobus:
:+1:

Ocramius:
Handled in #94. Thanks @Tobion and @kamermans!

ve3:
Thanks for the update! I almost change to use other cache library instead because I have no other choice than file system cache.
However I'm not sure that do you for get version number in **\Doctrine\Common\Cache\Version.php**? It is still '1.4.0-DEV'.

Ocramius:
Urgh, the version should indeed be bumped, but I see we didn't do it since 1.3.x :-\

PR #64, opened 2015-02-16
Reduce memory usage in SQLite3 cache

llaville:
Fetching data as array [1](http://www.php.net/manual/en/sqlite3result.fetcharray.php) use SQLITE3_BOTH mode by default.

Column number association is not used anywhere in the SQLite3 cache driver, but only column name.

On large data, we can improved a bit the memory usage.

Don't forget also the https://github.com/doctrine/cache/pull/62 that is really necessary to prove that this driver is operational.
I recommand also to fixed all others binding values to avoid confusion and error.

### DIFF
--- a/lib/Doctrine/Common/Cache/SQLite3Cache.php
+++ b/lib/Doctrine/Common/Cache/SQLite3Cache.php
@@ -179,7 +179,7 @@ class SQLite3Cache extends CacheProvider
 
         $statement->bindValue(':id', $id, SQLITE3_TEXT);
 
-        $item = $statement->execute()->fetchArray();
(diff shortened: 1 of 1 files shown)


Ocramius:
@llaville consider PR-ing an alternate fix for #62, as I'm not merging without a test case first.

Ocramius:
Merged, thanks!

llaville:
@Ocramius Thanks for quick merging. About #62, if PR author is not able to provide a test case, I may perharps (if everyone is agree) to propose one !

Ocramius:
@llaville yes please, just open a new PR with your test and his commit ;-)

PR #65, opened 2015-02-16
Unit test to solve original PR#62

llaville:
Here is a unit to demonstrate issue that truncate data.

1 - I've took the SQLite3Cache as object sample, but we can pick any other objects.
2 - I've fixed the `tearDown()` method, to close the sqlite3 DB instance before to remove file (or we got denied permission)

Without fix, you should get something like 

    1) Doctrine\Tests\Common\Cache\SQLite3Test::testFetchSingle
    Unexpected exception has been raised. unserialize(): Error at offset 47 of 51 bytes

### DIFF
--- a/tests/Doctrine/Tests/Common/Cache/SQLite3CacheTest.php
+++ b/tests/Doctrine/Tests/Common/Cache/SQLite3CacheTest.php
@@ -22,12 +22,35 @@ class SQLite3Test extends CacheTest
 
     protected function tearDown()
     {
+        $this->sqlite = null;  // DB must be closed before
         unlink($this->file);
     }
 
     public function testGetStats()
     {
         $this->assertNull($this->_getCacheDriver()->getStats());
+    }
+
+    public function testFetchSingle()
+    {
+        $id = uniqid('sqlite3_id_');
+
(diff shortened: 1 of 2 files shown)


Ocramius:
:+1:

tests/Doctrine/Tests/Common/Cache/SQLite3CacheTest.php
```diff
@@ -30,6 +31,28 @@ public function testGetStats()
...
+
+        $data = $this->_getCacheDriver();
+
+        $this->_getCacheDriver()->save($id, $data, 30);
+
+        try {
+            $actual = $this->_getCacheDriver()->fetch($id);
+
+        } catch (\PHPUnit_Framework_Error $e) {
+            $this->fail('Unexpected exception has been raised. ' . $e->getMessage());
```
stof: Don't catch ``\PHPUnit_Framework_Error`` to make it fail again. PHPUnit already mark the testsuite as failed in case an error occurs. You don't need to do it yourselves.

Ocramius: @stof fixed on merge, FYI

PR #66, opened 2015-03-09
Added support for DynamoDB as Cache Provider

inouet:


composer.json
```diff
@@ -18,7 +18,8 @@
     "require-dev": {
         "phpunit/phpunit":         ">=3.7",
         "satooshi/php-coveralls":  "~0.6",
-        "predis/predis":           "~1.0"
+        "predis/predis":           "~1.0",
+        "aws/aws-sdk-php":         "2.*"
```
Ocramius: Please use a mock instead

inouet:
By using DynamoDB Local, we will be able to test without the AWS resources.
(DynamoDB Local is a small client-side database and server that mimics the DynamoDB service.)

If you're using a Mac, you can install in this way.

$ brew install dynamodb-local
$ /usr/local/bin/dynamodb-local

Still, do I need a Mock?

Ocramius:
Yes, it really should be mocked out IMO :-\

Ocramius:
Won't fix, as this introduces:

 - a dependency to the AWS SDK (not wanted)
 - coupling with proprietary databases (not really wanted, although acceptable)
 - slow tests (should be mocked)

I suggest simply moving the adapter to a separate library that is AWS+doctrine specific.

PR #68, opened 2015-03-30
Reduced number of subdirectories generated by FileCache

Crozin:
Currently FileCache implementation generates 32 subdirectories for each cache entry. This is inefficient (FS), useless (every single file is going to be stored in its own subdirectory) and leads to quite hard to debug/detect bugs caused by max filepath length (260 chars on Windows).

### DIFF
--- a/lib/Doctrine/Common/Cache/FileCache.php
+++ b/lib/Doctrine/Common/Cache/FileCache.php
@@ -111,7 +111,7 @@ abstract class FileCache extends CacheProvider
     {
         return $this->directory
             . DIRECTORY_SEPARATOR
-            . implode(str_split(hash('sha256', $id), 2), DIRECTORY_SEPARATOR)
+            . implode(str_split(substr(hash('sha256', $id), 0, 8), 2), DIRECTORY_SEPARATOR)
             . DIRECTORY_SEPARATOR
             . preg_replace($this->disallowedCharacterPatterns, $this->replacementCharacters, $id)
             . $this->extension;
--- a/tests/Doctrine/Tests/Common/Cache/FileCacheTest.php
+++ b/tests/Doctrine/Tests/Common/Cache/FileCacheTest.php
@@ -94,15 +94,10 @@ class FileCacheTest extends \Doctrine\Tests\DoctrineTestCase
 
     public function testFilenameShouldCreateThePathWithFourSubDirectories()
     {
-        $cache          = $this->driver;
-        $method         = new \ReflectionMethod($cache, 'getFilename');
-        $key            = 'item-key';
-        $expectedDir    = array(
-            '84', 'e0', 'e2', 'e8', '93', 'fe', 'bb', '73', '7a', '0f', 'ee',
-            '0c', '89', 'd5', '3f', '4b', 'b7', 'fc', 'b4', '4c', '57', 'cd',
-            'f3', 'd3', '2c', 'e7', '36', '3f', '5d', '59', '77', '60'
-        );
-        $expectedDir    = implode(DIRECTORY_SEPARATOR, $expectedDir);
+        $cache       = $this->driver;
+        $method      = new \ReflectionMethod($cache, 'getFilename');
+        $key         = 'item-key';
+        $expectedDir = implode(DIRECTORY_SEPARATOR, array('84', 'e0', 'e2', 'e8'));
 
         $method->setAccessible(true);
 


Ocramius:
Few notes:

 - file modes were changed (to be reverted)
 - I don't see the problem with windows, since a `sha256` is actually quite short
 - this strategy increases collisions by a lot

Crozin:
> I don't see the problem with windows, since a sha256 is actually quite short

64 characters from hash and another 31 for directory separators gives total of 95 extra characters to filepath. That's over 1/3 off allowed length. Filename itself is also quite long. The very reason I've propsed this pull request is because I myself encountered this issue.

> this strategy increases collisions by a lot

Well... it does not. The collision will occur only if we use the very same `$id` for two cache files. And because `HASH($id) = HASH($id)` is always true, the length of returned hash is irrelevant. It's only used to reduce the number of files in single directory. Having 100 millions cache files, each with unique ID we'll end up with:
- 100 millions cache files in single directory unless we use some mechanism to evenly distribute them in some subdirectories, what we do,
- 100 millions 8-characters-long directories each containing another subdirectory with subdirectory with subdirectory with actual cache file if we use the old mechanism fixed in #16,
- 256 subdirectories each containg another 256 subdirecotries each containg another 256 subdirecotries each containg each containing another few direcotries each containing a chain of ~28 directories with a single subdirectory and at the very end of this chain a single cache file if we use mechanism propsed in #16,
- the very same result as above but without useless chain of ~28 "empty" directories and a few cache files in bottom-most subdirectory instead of one if we use mechanism propsed in this pull request.

> already fixed in #16

#16 fixed the issue of having ridiculous number of directories (see second point from the list above) just to bring us the issue of ~28 "empty" subdirectories causing problems related to maximum filepath length

Ocramius:
Reopened: ACK

zeroedin-bill:
Is there any reason to have nested subdirectories at all?

Why not just do this? I don't see this colliding ever...
```php
         return $this->directory
             . DIRECTORY_SEPARATOR
             . substr(hash('sha256', $id), 0, 16)
             . DIRECTORY_SEPARATOR
             . preg_replace($this->disallowedCharacterPatterns, $this->replacementCharacters, $id)
             . $this->extension;
```

Ocramius:
> Is there any reason to have nested subdirectories at all?

Probably not

guilhermeblanco:
@Ocramius @zeroedin-bill there is... depending on what is being cached, you may end up having more than 16k files in a single directory, causing new files to be ignored on write.

I'd suggest we break down in 2 levels of directories only.

zeroedin-bill:
@guilhermeblanco @Ocramius any sane system would encounter inode exhaustion *long* before a single directory got to 16k entries...

Ocramius:
Yes, but it has been demonstrated (by multiple PRs) that people don't know when to stop using a filesystem cache :(

zeroedin-bill:
@Ocramius @guilhermeblanco - how about using the first 16, and increasing the size of the path segment with each increase in depth? Statistically, this would exhaust much, much later than using a single directory for the first 16 characters.

example:
```php
$hash = substr(hash('sha256', 'foo'), 0, 16);
$path = substr($hash, 0, 2) . DS . substr($hash, 2, 4) . DS . substr($hash, 6) . $filename;
$path => 2c/26b4/6b68ffc68f/filename.php
```

pjcdawkins:
:+1:

trsteel88:
This is causing quite a few problems. I have a cache directory that takes over an hour to delete because there are so many directories. Are we able to get this merged?

zeroedin-bill:
@Ocramius @guilhermeblanco we need to fix this - If it's here, people will use it, and if it sucks, it's our fault...

Ocramius:
Yep, but the current fix isn't finalized

trsteel88:
@Ocramius, do you have any idea of when this might be fixed? Is there an open ticket elsewhere we can monitor? Considering forking temporarily if it is going to take a while.

enumag:
@Ocramius What is the problem with this PR? Why did you say it isn't finalized?

mvrhov:
@Ocramius: This bug really is a problem and a complete WTF. Before that the files were in a single directory and ATM I'm unable to delete the annotations directory on linux machine, shared over network because the tree structure is too long. Can we shorten this on at most 2 levels deep please. Because paths like this are not amusing at all: N:\naviter\cloud\site\app\www\var\cache\dev\annotations\05\62\12\99\03\cb\83\17\6b\74\3a\e1\58\68\99\20\cb\65\5b\f4\4e\65\c2\bf\be\24\ff\c3\27\83\9a\67\[[C]Symfony-Bundle-TwigBundle-Controller-ExceptionController-[Annot]][1].doctrinecache.data

untitled

trsteel88:
I agree with max 2 levels (with 1 letter or number per directory). The cache id should probably be moved to the filename itself.

Any more than 1 char directory still creates too many directories.

trsteel88:
This is what I am thinking: https://github.com/trsteel88/cache/commit/342d4f60118859b8273d4fb0fa075f251ff7fecf

It limits the sub directories name to 0-9 so max number of directories will be 3^9. The hashed id is also moved to the filename to ensure a unique path.

pjcdawkins:
Looks like anyone interested should move over to #94 - that PR is progressing more than this one

Tobion:
This can be closed as #94 has been merged.

Ocramius:
Indeed, thanks @Tobion!

PR #69, opened 2015-04-07
Improved PhpFileCache performance by removing useless stat calls

Ocramius:
Not sure if it's relevant, but this makes the `PhpFileCache` actually usable in a context where OpCache is enabled.

### DIFF
--- a/lib/Doctrine/Common/Cache/PhpFileCache.php
+++ b/lib/Doctrine/Common/Cache/PhpFileCache.php
@@ -42,6 +42,7 @@ class PhpFileCache extends FileCache
      */
     protected function doFetch($id)
     {
+        // note: error suppression is still faster than `file_exists`, `is_file` and `is_readable`
         $value = @include $this->getFilename($id);
 
         if (! isset($value['lifetime'])) {
@@ -60,6 +61,7 @@ class PhpFileCache extends FileCache
      */
     protected function doContains($id)
     {
+        // note: error suppression is still faster than `file_exists`, `is_file` and `is_readable`
         $value = @include $this->getFilename($id);
 
         if (! isset($value['lifetime'])) {
--- a/lib/Doctrine/Common/Cache/PhpFileCache.php
+++ b/lib/Doctrine/Common/Cache/PhpFileCache.php
@@ -42,10 +42,9 @@ class PhpFileCache extends FileCache
      */
     protected function doFetch($id)
     {
-        // note: error suppression is still faster than `file_exists`, `is_file` and `is_readable`
-        $value = @include $this->getFilename($id);
+        $value = $this->includeFileForId($id);
 
-        if (! isset($value['lifetime'])) {
+        if (! $value) {
             return false;
         }
 
@@ -61,10 +60,9 @@ class PhpFileCache extends FileCache
      */
     protected function doContains($id)
     {
-        // note: error suppression is still faster than `file_exists`, `is_file` and `is_readable`
-        $value = @include $this->getFilename($id);
+        $value = $this->includeFileForId($id);
 
-        if (! isset($value['lifetime'])) {
+        if (! $value) {
             return false;
         }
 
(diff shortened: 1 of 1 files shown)


lib/Doctrine/Common/Cache/PhpFileCache.php
```diff
@@ -42,14 +42,12 @@ public function __construct($directory, $extension = self::EXTENSION)
      */
     protected function doFetch($id)
     {
-        $filename = $this->getFilename($id);
+        $value = @include $this->getFilename($id);
```
Ocramius: yes, yes, I know there's an error suppression here.

We can't do otherwise, because this is the only way to avoid `is_file` and `is_readable` calls.

zeroedin-bill: *gasp*

Makes sense, but you might want to add a comment to these methods to the effect of your comment here. It's not immediately obvious from looking at the code.

Ocramius: will do

zeroedin-bill:
:+1:

lib/Doctrine/Common/Cache/PhpFileCache.php
```diff
@@ -42,14 +42,13 @@ public function __construct($directory, $extension = self::EXTENSION)
      */
     protected function doFetch($id)
     {
-        $filename = $this->getFilename($id);
+        // note: error suppression is still faster than `file_exists`, `is_file` and `is_readable`
+        $value = @include $this->getFilename($id);
```
stof: The error suppression should cover only the inclusion, not the filename computation IMO

Ocramius: Good point!

stof:
:+1:

Rican7:
Oooooo, [any inspiration here](https://twitter.com/trevorsuarez/status/570379383006846977)? ;)

Ocramius:
@Rican7 I probably misunderstood your first tweet, but I guess we're on the same line now :)

Rican7:
:smiley: :+1:

lib/Doctrine/Common/Cache/PhpFileCache.php
```diff
@@ -106,4 +98,23 @@ protected function doSave($id, $data, $lifeTime = 0)
...
+     * @param string $id
+     *
+     * @return array|false
+     */
+    private function includeFileForId($id)
+    {
+        $fileName = $this->getFilename($id);
+
+        // note: error suppression is still faster than `file_exists`, `is_file` and `is_readable`
+        $value = @include $fileName;
```
Rican7: Oh wow, I didn't expect this error suppression to be faster than using a temporary error handler either: http://3v4l.org/RFlvO. Its faster in almost every version. Interesting.

Ocramius: @Rican7 error handler replacement involves a lot of method calls - not efficient at all.

Rican7: Yea, I guess so. Too bad. Its not like this will actually be any better in PHP 7 with exceptions either (for actual errors, not warnings) since exceptions aren't cheap either.

urodoz: Since this pull requests, a php warning is triggered each time we create a new file cache.

Sample of message:

```bash
include(): Failed opening '/.../app/cache/prod/store_cache/d9/ec/a2/...[sample].storeschema.php
```
Removing the warnings with the php.ini file is an option, but can this Cache be configurable to enable or disable this error suppression?

Ocramius: @urodoz fix your xdebug settings - we have an error suppression exactly for this

urodoz: Thanks!

PR #70, opened 2015-04-10
Handle circular references in PhpFileCache

xhuberty:
In our homemade framework we could not only used PhpFileCache because our class (which is almost the same as yours) does not handle circular references.

Here is the solution we decided to implement. With this new implementation of this class you could used PhpFileCache instead of FileCache in the whole project.

### DIFF
--- a/lib/Doctrine/Common/Cache/PhpFileCache.php
+++ b/lib/Doctrine/Common/Cache/PhpFileCache.php
@@ -86,14 +86,6 @@ class PhpFileCache extends FileCache
             $lifeTime = time() + $lifeTime;
         }
 
-        if (is_object($data) && ! method_exists($data, '__set_state')) {
-            throw new \InvalidArgumentException(
-                "Invalid argument given, PhpFileCache only allows objects that implement __set_state() " .
-                "and fully support var_export(). You can use the FilesystemCache to save arbitrary object " .
-                "graphs using serialize()/deserialize()."
-            );
-        }
-
         $filename  = $this->getFilename($id);
 
         $value = array(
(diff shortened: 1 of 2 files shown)


xhuberty:
I had new phpunit test that show that the current implementation of PhpFileCache class can't handle circular references.

Furthermore the new implementation of PhpFileCache have a fallback if the object passed doesn't implement __set_state().

Ocramius:
This seems naive to me: what happens if the saved item is an array of objects, for example?

tests/Doctrine/Tests/Common/Cache/PhpFileCacheTest.php
```diff
@@ -65,12 +65,25 @@ public function testImplementsSetState()
         $this->assertTrue($cache->contains('test_set_state'));
     }
 
-    public function testNotImplementsSetState()
+    /*public function testNotImplementsSetState()
```
stof: this should not be commented. If it is not relevant anymore, it should be removed. Keeping track of the history is done using Git, not using code comments (which suck at it anyway).

tests/Doctrine/Tests/Common/Cache/PhpFileCacheTest.php
```diff
@@ -100,6 +113,10 @@ public function __construct($value)
         $this->value = $value;
     }
 
+    public function setValue($value)
+    {
+     $this->value = $value;
```
stof: wrong indentation

lib/Doctrine/Common/Cache/PhpFileCache.php
```diff
@@ -101,8 +93,14 @@ protected function doSave($id, $data, $lifeTime = 0)
             'data'      => $data
         );
 
-        $value  = var_export($value, true);
-        $code   = sprintf('<?php return %s;', $value);
+        $serializeValue = serialize($value);
+
+        if (strpos($serializeValue, 'r:') !== false || (is_object($data) && !method_exists($data, '__set_state'))) {
```
Ocramius: I'd rather throw an exception, now that I think of it.

xhuberty: Before the modification I made the PhpFileCache was throwing an exeption when $data was an object without method __set_state(). 

The modification I made trigger when we can't var_export $data (redondant reference or method __set_state() not implemented) and serialize it.

beberlei:
This cache is explicty written for supporting non circular object graphs that have __set_state() methods. Changing this is a huge BC break. Please write a new cache driver.

lib/Doctrine/Common/Cache/PhpFileCache.php
beberlei: This code assumes only the root object needs __set_state(), however every child needs the method as well. its very dangerous to mix this logic here with two completely different serialization approaches.

Ocramius:
Closing as per @beberlei's review. Write a separate cache adapter instead (if needed)

PR #71, opened 2015-04-28
Fix PHPDoc for doFetch return type

oconnedk:
Changed confusing string return type for doFetch, which *can* return string, but will generally return an object - as defined in [Cache::fetch](https://github.com/doctrine/cache/blob/master/lib/Doctrine/Common/Cache/Cache.php#L55)


GrahamCampbell:
?

GrahamCampbell:
if `mixed` is used, we cannot add other types to it as well. It should JUST say `mixed`, or even better, actually list all the possible types.

Ocramius:
@GrahamCampbell this PR is actually fine as-is

PR #78, opened 2015-05-06
Fix various issues with PredisCache and RedisCache

zeroedin-bill:
Sorry for PR spam :laughing:

### DIFF
--- a/lib/Doctrine/Common/Cache/PredisCache.php
+++ b/lib/Doctrine/Common/Cache/PredisCache.php
@@ -36,7 +36,7 @@ class PredisCache extends CacheProvider
             return false;
         }
 
-        return $result;
+        return unserialize($result);
     }
 
     /**
@@ -46,7 +46,7 @@ class PredisCache extends CacheProvider
     {
         $fetchedItems = call_user_func_array(array($this->client, 'mget'), $keys);
 
-        return array_filter(array_combine($keys, $fetchedItems));
+        return array_filter(array_combine($keys, array_map('unserialize', $fetchedItems)));
     }
     /**
      * {@inheritdoc}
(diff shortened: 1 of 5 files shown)


lib/Doctrine/Common/Cache/PredisCache.php
```diff
@@ -61,6 +61,7 @@ protected function doContains($id)
      */
     protected function doSave($id, $data, $lifeTime = 0)
     {
+        $data = serialize($data);
```
Ocramius: Empty line after zis

Ocramius:
:+1:

tests/Doctrine/Tests/Common/Cache/CacheTest.php
```diff
@@ -307,6 +309,52 @@ public function falseCastedValuesProvider()
...
+        $cache->save("obj1", $obj1);
+        $obj2 = new \stdClass();
+        $obj2->bar = "baz";
+        $cache->save("obj2", $obj2);
+
+        $fetched = $cache->fetchMultiple(array("obj1", "obj2"));
+        $this->assertInstanceOf("stdClass", $fetched["obj1"]);
+        $this->assertInstanceOf("stdClass", $fetched["obj2"]);
+        $this->assertEquals("bar", $fetched["obj1"]->foo);
+        $this->assertEquals("baz", $fetched["obj2"]->bar);
```
vvondra: I'd test cache misses as well

razvanphp:
Can you please tag this into a new minor version? It's been up there for more than a month.
We are using the master because of this (serious) bug.
Thank you!

davedevelopment:
I just got stung by this, could probably have done with something in the changelog? A hot cache that had already serialized data in was returning incorrect objects etc, when application code was expecting serialized data. 

Anyway, not to worry, but if anyone else comes confused about problems going from 1.4.1 to 1.4.2, it could be this.

PR #79, opened 2015-05-13
Add stats on hits/misses in redis caches

romainneutron:


Ocramius:
@romainneutron can you test that those keys are not-null now? Basically an extension to https://github.com/doctrine/cache/blob/c2ab51e8f3c4a8170ca91b702e3789881dd2a6c2/tests/Doctrine/Tests/Common/Cache/CacheTest.php#L254

romainneutron:
I've added tests, is it what you expected?

Ocramius:
Looking good, thanks! Waiting for travis.

Ocramius:
:+1:

PR #80, opened 2015-06-03
Enhancement - Improve Flushable/Clearable interface documentation

Rican7:
I was browsing the different common interfaces in Doctrine Cache and happened upon both the `Flushable` and `Clearable` interfaces and for the life of me couldn't figure out the difference between the two.

[After a quick conversation with @Ocramius][ocramius-twitter-cache-thread], he helped me understand the difference. I figured I'd improve the docs here to help clarify their intents so that others wouldn't get into the same confusion and hopefully help improve discoverability.

[ocramius-twitter-cache-thread]: https://twitter.com/trevorsuarez/status/606133921542217728

### DIFF
--- a/lib/Doctrine/Common/Cache/ClearableCache.php
+++ b/lib/Doctrine/Common/Cache/ClearableCache.php
@@ -23,7 +23,7 @@ namespace Doctrine\Common\Cache;
  * Interface for cache that can be flushed.
  *
  * Intended to be used for partial clearing of a cache namespace. For a more
- * global "flushing", see {@see FlushableCache}
+ * global "flushing", see {@see FlushableCache}.
  *
  * @link   www.doctrine-project.org
  * @since  1.4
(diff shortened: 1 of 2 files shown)


lib/Doctrine/Common/Cache/ClearableCache.php
```diff
@@ -29,7 +32,7 @@
 interface ClearableCache
 {
     /**
-     * Deletes all cache entries.
+     * Deletes all cache entries in the current cache namespace.
```
Ocramius: Note that namespaces are a specific feature of some cache types, not necessarily of a `ClearableCache`

Rican7: Hmm, so if a cache type/adapter doesn't support namespaces, this is pretty much just the `Flushable` cache?

Rican7: #bump

Rican7: #bump

andrerom:
Was just browsing the same and could not get the difference either, +1 for some kind of cleanup. Maybe move flush method to NamespacedCache if the discussion above turns out to be true.

lib/Doctrine/Common/Cache/ClearableCache.php
Ocramius: Correct. Merging as-is btw: the clarification is ok.

PR #81, opened 2015-06-07
Added session-based cache provider

rtheunissen:
This branch adds a basic session-based cache provider. The use case I came across was to store an OAuth2 access token, where the default cache provider is the user session, but you can use a Doctrine Cache Provider instead. To make dependency injection easier, it would be nice to be able to accept a Doctrine Cache Provider, while preserving the default session-based storage.

### DIFF
--- a/lib/Doctrine/Common/Cache/SessionCache.php
+++ b/lib/Doctrine/Common/Cache/SessionCache.php
@@ -66,7 +66,11 @@ class SessionCache extends CacheProvider
      */
     protected function doContains($id)
     {
-        return isset($_SESSION[$id]);
+        if (!isset($_SESSION)) {
+            return false;
+        }
+
+        return isset($_SESSION[$id]) || array_key_exists($id, $_SESSION);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


rtheunissen:
Can someone please get back to me on this?

zeroedin-bill:
@rtheunissen There are a lot of things that make a PHP session unsuitable as a generic cache. This cache wouldn't be very useful as a metadata cache or generic cache. For one thing, you'd have blocking issues due to session write locking. You'd also have to serialize and unserialize the entire cache every time you open and close the session. 

The cache providers in this library are targeted toward use as a generic cache, or at least as a fast metadata cache. 

I understand your specific use case, but I think that maybe this is not the best candidate for inclusion as part of the cache library.

Ocramius:
The idea is neat, but based on the broken concept that you can use your PHP session as a cache.

PHP sessions are a mess, especially if based on `ext/session`. Consider using a by-ref-array cache instead, so you can use whatever array you prefer.

You can simply use the code of the `ArrayCache` for doing so, but I don't think it's a good use-case for `doctrine/cache`.

Closing, sorry.

PR #82, opened 2015-06-29
Fix race condition in FileCache-based drivers.

krixon:
This fixes a race condition which occurs when the directory was created between the calls to `is_dir` and `mkdir` by checking `is_dir` again after `mkdir`. If `mkdir` failed but `is_dir` return true anyway, there is no problem.


krixon:
No test for this as it's pretty impossible to set up a failure case :)

Ocramius:
:+1:

PR #84, opened 2015-08-07
Enhancement - Improve CacheProvider inheritance

mkovel:


kimhemsoe:
We really want to avoid making attributes protected. It is just another API we need to maintain.

Any specific reason you want this ?
If you really need something special there is an interface you can use and implement everything you want.

Ocramius:
Sorry, but unless there is a strong need for this change inside the package itself, you shouldn't rely on inheriting from `CacheProvider` if you need these properties.

ja1cap:
@Ocramius @kimhemsoe we have implemented saveMultiple method
```php
  /**
   *
   * @param array [id] $saveItemStack
   * @param int $lifeTime
   *
   * @return array
   * @throws \Exception
   */
  public function saveMultiple( array $saveItemsStack, $lifeTime = 0 ) {

    $ids                  = array_keys( $saveItemsStack );
    $namespacedIds        = array_map( array( $this, 'getNamespacedId' ), $ids );
    $namespacedItemsStack = array_combine( $namespacedIds, $saveItemsStack );

    $saveResult = $this->getRedis()->mSet( $namespacedItemsStack );

    if ( $saveResult && $lifeTime != 0 ) {
      foreach ( $namespacedIds as $namespacedId ) {
        $this->getRedis()->expire( $namespacedId, $lifeTime );
      }
    }

    return array_fill_keys( $ids, $saveResult );
  }
```

In this method we'd like to use the same id (key that is used to store value in cache), maybe you can make getNamespacedId method final protected to avoid problems that u've described 
```php
    /**
     * Prefixes the passed id with the configured namespace value.
     *
     * @param string $id The id to namespace.
     *
     * @return string The namespaced id.
     */
    final protected function getNamespacedId($id)
    {
        $namespaceVersion  = $this->getNamespaceVersion();

        return $this->namespace."[".$id."][".$namespaceVersion."]";
    }
```

Thanks

mkovel:
Thank @ja1cap that's what I wanted to say

PR #85, opened 2015-08-10
Ensure users do not mix caches on POSIX filesystems

jeskew:


lib/Doctrine/Common/Cache/FileCache.php
```diff
@@ -158,16 +144,51 @@ protected function doGetStats()
...
+        if ( ! $this->createPathIfNeeded($directory, 0777)) {
+            throw new \InvalidArgumentException(sprintf(
+                'The directory "%s" does not exist and could not be created.',
+                $directory
+            ));
+        }
+
+        if (function_exists('posix_geteuid')) {
+            $userData = posix_getpwuid(posix_geteuid());
+            $directory .= DIRECTORY_SEPARATOR
```
Ocramius: This change seems to create parallel cache dirs (one per user). Looks incorrect to me, although I understand why you tried to go in this direction.
I simply would suggest not sharing a cache dir across different users (seems like a bad idea in first place)

lib/Doctrine/Common/Cache/FileCache.php
```diff
@@ -158,16 +144,51 @@ protected function doGetStats()
...
+                . $userData['name'];
+
+            if ( ! $this->createPathIfNeeded($directory)
+                || fileowner($directory) !== posix_geteuid()
```
stof: this makes the system incompatible with filesystem ACLs AFAICT

jeskew: It would. It's mooted by [the core team's blog post on umasks](http://www.doctrine-project.org/2015/08/31/security_misconfiguration_vulnerability_in_various_doctrine_projects.html), so I'll go ahead and close it out.

jeskew:
Thanks for taking a look at this. Enforcing a proper umask is a more elegant way to accomplish what this PR was trying to do, so I don't think it's still relevant.

PR #86, opened 2015-08-18
Enhancement - Improve CacheProvider inheritance: Version 2

mkovel:
I have considered your comments, I would like to offer you another variant


Ocramius:
I would still not extend from this class if you need to use its internal components. Just rewrite from scratch.

mkovel:
I don't understand what you don't like protected + final ?

Ocramius:
The problem is that this approach enforces us to treat this API as a possible source of BC breaks, whereas we simply want to be able to rewrite or scrap it at any time.

Copying over the logic is perfectly fine.

mkovel:
ok, i understood your opinion. tnx

deeky666:
Closing as per discussion.

PR #88, opened 2015-09-03
Add an OpenSSL-based encrypting cache decorator

jeskew:
In case you want to store something sensitive in a cache and can live with the performance hit from OpenSSL encryption.

### DIFF
--- a/lib/Doctrine/Common/Cache/EncryptingCacheDecorator.php
+++ b/lib/Doctrine/Common/Cache/EncryptingCacheDecorator.php
@@ -1,0 +1,94 @@
+<?php
+namespace Doctrine\Common\Cache;
+
+use InvalidArgumentException as IAE;
+use RuntimeException as RE;
+
+class EncryptingCacheDecorator implements Cache
+{
+    /** @var Cache */
+    private $decorated;
+    /** @var resource */
+    private $publicKey;
+    /** @var resource */
+    private $privateKey;
+
+    /**
+     * @param Cache $decorated
+     * @param resource $publicKey
+     * @param resource $privateKey
+     *
+     * @throws RE If OpenSSL extension not installed.
+     * @throws IAE If either key is not an OpenSSL Key resource.
+     */
+    public function __construct(Cache $decorated, $publicKey, $privateKey)
+    {
+        if (!extension_loaded('openssl')) {
+            throw new RE('The OpenSSL extension is required to use the'
+                . __CLASS__ . '. Please install OpenSSL to use this feature.');
+        }
+
+        foreach (array($publicKey, $privateKey) as $key) {
+            if (!is_resource($key) || 'OpenSSL key' !== get_resource_type($key)) {
+                throw new IAE('Keys passed to ' . __CLASS__ . 'must be OpenSSL'
+                    . ' key resources.');
+            }
+        }
+
+        $this->decorated = $decorated;
+        $this->publicKey = $publicKey;
+        $this->privateKey = $privateKey;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function fetch($id)
+    {
+        $stored = $this->decorated->fetch($id);
+        if (isset($stored['data'])) {
+            openssl_private_decrypt($stored['data'], $decrypted, $this->privateKey);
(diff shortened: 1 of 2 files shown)


lib/Doctrine/Common/Cache/EncryptingCacheDecorator.php
```diff
@@ -0,0 +1,94 @@
+<?php
+namespace Doctrine\Common\Cache;
+
+use InvalidArgumentException as IAE;
+use RuntimeException as RE;
```
stof: please don't alias exceptions like this. It is quite unreadable

lib/Doctrine/Common/Cache/EncryptingCacheDecorator.php
```diff
@@ -0,0 +1,94 @@
...
+     */
+    public function __construct(Cache $decorated, $publicKey, $privateKey)
+    {
+        if (!extension_loaded('openssl')) {
+            throw new RE('The OpenSSL extension is required to use the'
+                . __CLASS__ . '. Please install OpenSSL to use this feature.');
+        }
+
+        foreach (array($publicKey, $privateKey) as $key) {
+            if (!is_resource($key) || 'OpenSSL key' !== get_resource_type($key)) {
```
stof: OpenSSL functions also accept a path to a pem certificate file when calling encryption and decryption functions. Supporting it here would make it much simpler to use the class (especially when using a DI container to manage the instantiation: it is much easier to make it pass a string to the class than to make it build resources). It would also make it much easier to handle the lifecycle of the class (as you would not need to free the key resource anymore to avoid leaking it, which is currently quite hard to do)

Ocramius:
I disagree with these bits:

 > In case you want to store

A cache is not really a storage :|

 > something sensitive in a cache

A cache is to be treated as something sensitive anyway, as if it was part of the address space of the app (although it almost never is). Everything in a cache is sensitive to any app, as it can modify the behavior of the app greatly, if touched.

In general, this may be one approach, but I'd rather not have it in the `doctrine/cache` repository, as 99.9% of the users would simply abuse it, misunderstanding how a cache should be used in first place.

jeskew:
@Ocramius any cache that writes to disk would be covered by corporate or government policies requiring the encryption of data at rest. That would include caches using files, Redis, or any kind of persistent database. In those cases, your cache is far more sensitive than the source code of your application, at least from a compliance standpoint.

Ocramius:
>  government policies requiring the encryption of data at rest.

Meh, I yet got to see such a thing happening, and in that case I'd expect them to be using `<proprietary-encryption-system-that-all-of-the-open-source-community-discourages decision-taken-by-uninformed-manager="true"/>` (yes, sorry, this is a rant, I just felt it was necessary).

 >  That would include caches using files, Redis, or any kind of persistent database. In those cases, your cache is far more sensitive than the source code of your application, at least from a compliance standpoint.

Yeah, but the fact is that this sort of compliance is very (VERY) use-case specific, and doesn't fit this general-purpose library.

If something like this patch would land into `doctrine/cache` we'd be riddled with dozens of different patches about different supported crypto algos (symmetric, asymmetric, custom, etc), as well as major security maintenance load that we don't really want to take on (we are not crypto people within the doctrine orga).

Ocramius:
Discussed it internally:

 - we don't have the resources to maintain crypto functionality
 - since decoration is possible, decoration can exist in a separate repo

Closing therefore, sorry @jeskew

jeskew:
That makes sense. Since this PR is just a cache decorator, it works just as well as a separate package. 

(To your other point: my understanding is that encrypt-data-at-rest policies are not uncommon, nor is the use of open source crypto by corporations and governmental entities. The NSA might be rolling its own cryptographic primitives, but I doubt the Department of Education is.)

Ocramius:
@jeskew sorry, working in a govt agency made me kinda acid on their choices about crypto :-)

jeskew:
@Ocramius lol no worries. One last thing: does Doctrine have a trademark policy? I'll probably use the word "doctrine" in the other repo's name if it's permitted.

Ocramius:
@jeskew we're pretty much MIT all the things

PR #89, opened 2015-09-09
Fix RedisCache::fetchMultiple if keys array is empty

Pouix:
If provided with an empty array, Redis mget returns false and array_combine expects parameter 2 to be array.
Therefore we need to check if $keys is empty and return an empty array.

### DIFF
--- a/lib/Doctrine/Common/Cache/RedisCache.php
+++ b/lib/Doctrine/Common/Cache/RedisCache.php
@@ -71,6 +71,9 @@ class RedisCache extends CacheProvider
      */
     protected function doFetchMultiple(array $keys)
     {
+        if (empty($keys)) {
+            return array();
+        }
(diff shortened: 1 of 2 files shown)


Ocramius:
Requires testing

fprochazka:
I've ran into this too. :+1:

Pouix:
With the small test I added on RedisCache, another matter shows up:
if the array $keys is empty, CacheProvider::fetchMultiple cashes because both parameters of array_combine should have at least 1 element (PHP 5.3).

So I removed my fix on RedisCache::fetchMultiple and put it into CacheProvider::fetchMultiple.
By doing so, both CacheProvider::fetchMultiple on PHP 5.3 and RedisCache::fetchMultiple on higher PHP version are fixed.

I close this PR and open a new one => https://github.com/doctrine/cache/pull/90

PR #90, opened 2015-09-10
Fix CacheProvider::fetchMultiple if keys array is empty

Pouix:
This is fixing some errors when we provide an empty array to CacheProvider::fetchMultiple :
- On PHP 5.3, Both parameters of array_combine should have at least 1 element.
- If we provide an empty array, Redis mget returns false and array_combine expects parameter 2 to be array.

### DIFF
--- a/lib/Doctrine/Common/Cache/CacheProvider.php
+++ b/lib/Doctrine/Common/Cache/CacheProvider.php
@@ -83,6 +83,9 @@ abstract class CacheProvider implements Cache, FlushableCache, ClearableCache, M
      */
     public function fetchMultiple(array $keys)
     {
+        if (empty($keys)) {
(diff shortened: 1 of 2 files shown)


lib/Doctrine/Common/Cache/CacheProvider.php
```diff
@@ -83,6 +83,9 @@ public function fetch($id)
      */
     public function fetchMultiple(array $keys)
     {
+        if (empty($keys)) {
+            return array();
+        }
         // note: the array_combine() is in place to keep an association between our $keys and the $namespacedKeys
```
zeroedin-bill: insert a newline here please

zeroedin-bill:
:+1:

PR #91, opened 2015-09-12
No need to save the default namespace version

Tobion:
1. The namepace version is increased in deleteAll() and also saved there. So when the version is still the default (1) there is no point in persisting that. This way we save space and unneeded workload.

2. do not update the cached namespace version when persistence failed. Otherwise all data written after a failed delete, will not be available later on.

### DIFF
--- a/tests/Doctrine/Tests/Common/Cache/CacheTest.php
+++ b/tests/Doctrine/Tests/Common/Cache/CacheTest.php
@@ -53,34 +53,6 @@ abstract class CacheTest extends \Doctrine\Tests\DoctrineTestCase
         );
     }
 
-    public function testFetchMultiWillFilterNonRequestedKeys()
-    {
-        /* @var $cache \Doctrine\Common\Cache\CacheProvider|\PHPUnit_Framework_MockObject_MockObject */
-        $cache = $this->getMockForAbstractClass(
-            'Doctrine\Common\Cache\CacheProvider',
-            array(),
-            '',
-            true,
-            true,
-            true,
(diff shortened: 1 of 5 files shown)


tests/Doctrine/Tests/Common/Cache/CacheTest.php
```diff
@@ -70,9 +70,9 @@ public function testFetchMultiWillFilterNonRequestedKeys()
             ->expects($this->once())
             ->method('doFetchMultiple')
             ->will($this->returnValue(array(
-                '[foo][]' => 'bar',
-                '[bar][]' => 'baz',
-                '[baz][]' => 'tab',
+                '[foo][1]' => 'bar',
```
Tobion: I just had to change that because `doFetch` wrongly returned null instead of false in the mocked tests.

Ocramius:
@Tobion could you cover each of your 2 points in the description of the PR with a separate test? (with assertions that include those messages?)

tests/Doctrine/Tests/Common/Cache/CacheTest.php
```diff
@@ -53,34 +53,6 @@ public function testFetchMulti()
         );
     }
 
-    public function testFetchMultiWillFilterNonRequestedKeys()
```
Tobion: doesnt make sense to run this test for each driver since its mocking

Tobion:
@Ocramius I added tests

Ocramius:
@Tobion thanks!

PR #93, opened 2015-09-18
Silence chmod() warnings

mpdude:
Also see https://github.com/doctrine/common/pull/383

### DIFF
--- a/lib/Doctrine/Common/Cache/PhpFileCache.php
+++ b/lib/Doctrine/Common/Cache/PhpFileCache.php
@@ -103,7 +103,7 @@ class PhpFileCache extends FileCache
         $code   = sprintf('<?php return %s;', $value);
 
         $ret = (file_put_contents($filename, $code) !== false);
-        chmod($filename, 0664);
+        @chmod($filename, 0664);
         return $ret;
     }
 }


lib/Doctrine/Common/Cache/PhpFileCache.php
```diff
@@ -103,7 +103,7 @@ protected function doSave($id, $data, $lifeTime = 0)
         $code   = sprintf('<?php return %s;', $value);
 
         $ret = (file_put_contents($filename, $code) !== false);
-        chmod($filename, 0664);
+        @chmod($filename, 0664);
```
Ocramius: What may be the outcome (later on) if there is a failure in executing this operation? Doesn't feel right to just silence it without handling it.

mpdude: I understand. Let's continue the discussion at https://github.com/doctrine/doctrine2/commit/caf30b889bb898620d843d1ec4940d01fa1d8877#commitcomment-13321422 (which is about the same issue and security problem) and come back later once we've got that sorted out.

Ocramius:
Closing here unless we find a different (workable) solution.

mpdude:
@ocramius As you just merged the similar fix in doctrine/common#383, do you want to think about this one again?

Ocramius:
@mpdude this can probably be reopened.

Ocramius:
@mpdude checked again: this affects only 1.3.x, and I won't backport to that branch, sorry.

PR #94, opened 2015-09-24
fix file naming based on cache key

Tobion:
Replaces #68 and #63

- Two hex chars to balance the tree are enough and is the best compromise. This is also used by twig now.
Files will be evenly distributed between 16^2 = 256 directories.
The old solution created unique directory tree for each file. For each file it created 32 directories! Totally unusable.

Using rawurlencode ensures that
- that the filename is already UNIQUE based on the cache key, so the extra dir balancing above is actually optional and could be configurable in the future
- only valid characters inside the filename (no NULL byte, no `/` and also not the other chars that are disallowed on Windows)
- the filename still resembles the cache key, i.e. valid chars stay as-is (good for debugging)
- faster than a `preg_replace`
- the `preg_replace` implementation was actually useless as it didn't prevent collisions at all, e.g. "-" and "__" would result in the same filename. This is probably also one of the reasons the hash above was introduced.


lib/Doctrine/Common/Cache/FileCache.php
```diff
@@ -126,9 +113,9 @@ protected function getFilename($id)
     {
         return $this->directory
             . DIRECTORY_SEPARATOR
-            . implode(str_split(hash('sha256', $id), 2), DIRECTORY_SEPARATOR)
+            . substr(hash('sha256', $id), 0, 2)
             . DIRECTORY_SEPARATOR
-            . preg_replace($this->disallowedCharacterPatterns, $this->replacementCharacters, $id)
+            . rawurlencode($id)
```
Ocramius: Indeed a much better solution than having a blacklist!

lib/Doctrine/Common/Cache/FileCache.php
```diff
@@ -126,9 +113,9 @@ protected function getFilename($id)
     {
         return $this->directory
             . DIRECTORY_SEPARATOR
-            . implode(str_split(hash('sha256', $id), 2), DIRECTORY_SEPARATOR)
+            . substr(hash('sha256', $id), 0, 2)
```
Ocramius: This is most probably enough indeed.
@guilhermeblanco @zeroedin-bill what is your opinion?

tests/Doctrine/Tests/Common/Cache/FileCacheTest.php
```diff
@@ -27,19 +27,19 @@ public function getProviderFileName()
...
+            array('key/3', 'key%2F3'),
+            array('key<4', 'key%3C4'),
+            array('key>5', 'key%3E5'),
+            array('key"6', 'key%226'),
+            array('key*7', 'key%2A7'),
+            array('key?8', 'key%3F8'),
+            array('key|9', 'key%7C9'),
+            array('key[10]', 'key%5B10%5D'),
+            array('keyä11', 'key%C3%A411'),
+            array('../key12', '..%2Fkey12'),
```
Ocramius: This is a potential security issue - the `.` character should be replaced.

lib/Doctrine/Common/Cache/FileCache.php
Tobion: With this we ensure 
- that the filename is already UNIQUE based on the cache key
- only valid characters inside the filename (no NULL byte, no `/` and also not the other chars that are disallowed on Windows)
- the filename still resembles the cache key, i.e. valid chars stay as-is (good for debugging)
- faster than a `preg_replace`
- the `preg_replace` implementation was actually useless as it didn't prevent collisions at all, e.g. "-" and "__" would result in the same filename. This is probably also one of the reasons the hash above was introduced.

Ocramius: > This is probably also one of the reasons the hash above was introduced.

Agreed, although the hash also guarantees some better distribution of the files across directories (otherwise, this cache has a hard limit imposed by the number of files per directory supported by the FS)

tests/Doctrine/Tests/Common/Cache/FileCacheTest.php
```diff
@@ -83,36 +84,29 @@ public function testFilenameCollision()
...
         );
         $expectedDir    = implode(DIRECTORY_SEPARATOR, $expectedDir);
 
         $method->setAccessible(true);
 
         $path       = $method->invoke($cache, $key);
-        $filename   = pathinfo($path, PATHINFO_FILENAME);
         $dirname    = pathinfo($path, PATHINFO_DIRNAME);
 
-        $this->assertEquals('item__key', $filename);
```
Ocramius: These two assertions should still be re-introduced: they would just assert that the dash was encoded.

lib/Doctrine/Common/Cache/FileCache.php
Tobion: Two hex chars to balance the tree are enough and is the best compromise. This is also used by twig now.
Files will be evenly distributed between 16^2 = 256 directories.

The old solution created unique directory tree for each file. For each file it created 32 directories! Totally unusable.

Ocramius: > Totally unusable.

Fully agree - we had an old running issue because of that (which can't be fixed in 1.4.x).

This approach puts us at a theoretical limit of 16776960 cache entries (in most cases). Should be enough.

Tobion: Todays filesystems do not even have this limitation of files per directory. So IMO this balancing should be opt-in. This way we could save a `hash` call per cache item and another inode that is wasted on most systems. Remember the cache should be fast, so there should not be unneeded workload for every item you save/fetch/delete etc.

Ocramius: Ah yes, it seems like ext4 (very common) is indeed not affected by that limitation, but tools start to crash and burn with over 10k files/dir

lib/Doctrine/Common/Cache/FileCache.php
Tobion: Yes as I that above, it would need to be configurable. But that is a different task.

tests/Doctrine/Tests/Common/Cache/FileCacheTest.php
Tobion: What security issue are you concerned about?

tests/Doctrine/Tests/Common/Cache/FileCacheTest.php
Tobion: The dash is not encoded anymore. And the encoding of the key is already covered in the other test. These are unrelated things that should not be mixed.

tests/Doctrine/Tests/Common/Cache/FileCacheTest.php
Ocramius: I'm mostly worried about directory traversal. While `.` is a perfectly valid file name, we may have crashes caused by clients producing cache keys that are just `.` or `..` (not allowed, afaik)

tests/Doctrine/Tests/Common/Cache/FileCacheTest.php
Ocramius: These tests dig deep into private/protected API to test specific behaviors. If you are reducing the units being tested, then there should be one test that covers the functional behavior (still an unit test, but integrates both these tested units)

tests/Doctrine/Tests/Common/Cache/FileCacheTest.php
Tobion: True, those files would not be possible to create.

Tobion: I encoded the dot to fix that.

tests/Doctrine/Tests/Common/Cache/FileCacheTest.php
Tobion: I don't know what you mean. But exactly this line is already tested via https://github.com/doctrine/cache/pull/94/files#diff-3bb49fafb3678735d131bb4e4d7b5348R44
There is no point in testing it again.

Tobion:
I think I need to encode uppercase letters on Windows as the file system is case-insensitive. Otherwise cache keys of different case would match on Windows.

lib/Doctrine/Common/Cache/FileCache.php
zeroedin-bill: Absolute :+1: - Given the very narrow range of acceptable uses for this cache, the likelihood that this cache will exhaust the filesystem with this patch *when used correctly* is extremely small.

tests/Doctrine/Tests/Common/Cache/FileCacheTest.php
Ocramius: @Tobion alright, if it's already covered then it is ok

lib/Doctrine/Common/Cache/FileCache.php
```diff
@@ -126,9 +114,9 @@ protected function getFilename($id)
...
             . DIRECTORY_SEPARATOR
-            . preg_replace($this->disallowedCharacterPatterns, $this->replacementCharacters, $id)
+            . str_replace('.', '%2E', rawurlencode($id))
```
Ocramius: As per discussion, needs encoding of uppercase ascii. Actually, to simplify, can't we have encoding of all chars?

matteodem:
What's the status of this? We're again having problems on systems with inodes being exhausted.

Ocramius:
@matteodem the only missing bit is the encoding of uppercase chars

Tobion:
@Ocramius I added a fix and test to ensure the cache key is case sensitive even on Windows filesystem

lib/Doctrine/Common/Cache/FileCache.php
```diff
@@ -124,11 +112,26 @@ public function getExtension()
      */
     protected function getFilename($id)
     {
+        if ('\\' === DIRECTORY_SEPARATOR) {
```
Ocramius: This is not really working, as the same behavior is expected also on linux/mac case-insensitive filesystems (yeah, macos is case-insensitive by default)

lib/Doctrine/Common/Cache/FileCache.php
```diff
@@ -124,11 +112,26 @@ public function getExtension()
...
+            $filename = preg_replace_callback('/[^a-z0-9~_\-]+?/', function ($matches) {
+                return '%' . bin2hex($matches[0]);
+            }, $id);
+        } else {
+            // Using percent-encoding we ensure that
+            // - the filename is unique based on the cache key, i.e. no collisions are possible
+            // - only valid characters inside the filename (no NULL byte, no / and also not the other chars that are disallowed on Windows)
+            // - the filename still resembles the cache key, i.e. valid chars stay as-is (good for debugging)
+            // Additionally, we encode the dot to prevent "." or ".." as filename.
+            $filename = str_replace('.', '%2E', rawurlencode($id));
```
Ocramius: This case is already handled by the previous conditional

lib/Doctrine/Common/Cache/FileCache.php
```diff
@@ -124,11 +112,26 @@ public function getExtension()
      */
     protected function getFilename($id)
     {
+        if ('\\' === DIRECTORY_SEPARATOR) {
+            // On Windows we also need to percent-encode uppercase characters as its filesystem is case-insensitive.
+            // The regex is non-greedy because we need to encode each character individually.
+            $filename = preg_replace_callback('/[^a-z0-9~_\-]+?/', function ($matches) {
```
Ocramius: What is the reason for excluding `~` and `_`?

tests/Doctrine/Tests/Common/Cache/CacheTest.php
```diff
@@ -29,6 +29,21 @@ public function testBasicCrudOperations($value)
...
+
+        $this->assertTrue($cache->save('key', 'value'));
+        $this->assertTrue($cache->contains('key'));
+        $this->assertSame('value', $cache->fetch('key'));
+
+        $this->assertFalse($cache->contains('KEY'));
+        $this->assertFalse($cache->fetch('KEY'));
+
+        $cache->delete('KEY');
+        $this->assertTrue($cache->contains('key', 'Deleting cache item with different case must not affect other cache item'));
```
Ocramius: This test passes also on case-insensitive operating systems because of the hash being different for different case keys: tricky :-\

lib/Doctrine/Common/Cache/FileCache.php
Tobion: to replicate the same behavior as when using rawurlencode

tests/Doctrine/Tests/Common/Cache/CacheTest.php
Tobion: i know but that cannot be avoided. there probably is key with different case that has the same first 2 chars in the hash. I I don't know a way to find this collision. So this is more a theoretical test that will be even more relavant in case we make the balancing hash logic configurable/optional.

lib/Doctrine/Common/Cache/FileCache.php
Tobion: ??? This is in an `else`

lib/Doctrine/Common/Cache/FileCache.php
Tobion: So what do you suggest to change? Can we detect case-insensitive fs in a sane manner? We could create a temporary file once in the constructor and check if we can access it case-insensitive. Do you prefer that?

tests/Doctrine/Tests/Common/Cache/CacheTest.php
Ocramius: > I don't know a way to find this collision

Same issue here, but collisions are really dangerous here.
I'd be happy to have a filesystem test that verifies that the file is saved in a specific location for two case-differing keys. This test case can still live here, as it serves other adapters as well

lib/Doctrine/Common/Cache/FileCache.php
Ocramius: @Tobion if we `bin2hex` all characters, then all these questions go away anyway

lib/Doctrine/Common/Cache/FileCache.php
Ocramius: Alright, although we can really `bin2hex` everything now.

lib/Doctrine/Common/Cache/FileCache.php
Ocramius: @Tobion just `bin2hex` every char in the key, IMO, and have that for both linux and windows. Sounds inefficient (probably is), but we can't try to discover case sensitivity anyway.

Tobion: The problem with `bin2hex` everything is that it increases the filename by a factor of two. Most filesystem have a limit of 255 chars for filename/path components. So when using a cache key with a length of >=128, it will create an error. And we lose the advantage of having the cache key = filename for non-special chars.
In this case we could just as well use the hash for the filename as well.

Ocramius: @Tobion or we simply use this first block for everything (instead of `rawurlencode()`)

Tobion:
@Ocramius I changed to using the hash for the filename. This is the fastest and most straight forward solution. I also added tests for cache id collisions and special ids chars for all cache providers.

This is ready from my side. Please merge soon as it's quite important for all people using file caches.

lib/Doctrine/Common/Cache/FileCache.php
```diff
@@ -124,11 +112,13 @@ public function getExtension()
      */
     protected function getFilename($id)
     {
+        $hash = hash('sha256', $id);
```
Ocramius: This has been discussed previously, and it is not really viable due to possible collisions anyway.

This approach can only work if the subclasses store the cache ID inside the stored file, and then check against it on retrieval. Probably the same is to be done also for deletions.

Tobion: What? Collisions on sha256? Please read http://stackoverflow.com/questions/4014090/is-it-safe-to-ignore-the-possibility-of-sha-collisions-in-practice for example.

> This approach can only work if the subclasses store the cache ID inside the stored file, and then check against it on retrieval. Probably the same is to be done also for deletions.

No idea what you are talking about.

Ocramius: I realize that the likeliness of a collision is pretty much zero, but since there is a chance, there must be a mechanism to check integrity.

 > No idea what you are talking about.

For example, https://github.com/doctrine/cache/blob/1c3483678ab767c3e21d3a96d6dd1006b327d425/lib/Doctrine/Common/Cache/FilesystemCache.php#L106 would have to save the key as part of the data, in order to check the key after retrieval.

Tobion: I wouldn't add a performance penalty to every cache for a chance of a collision that is so low that the likelihood of the world ending it higher. Even if a collision occurs, it doesn't matter. Its cache data, not nuclear weapon software.

Ocramius: You can do this sort of assumption in your own technical domain, but not in a general purpose lib.

Tobion: I implemented another versiong using bin2hex by default and fallback to the hash if id would be too long. Also I fixed the naming if id is empty, where bin2hex wouldn't work as well.

kcivey:
Is this really not going to be fixed until 1.5.0? These sub^32-directories are eating up an insane number of inodes for nothing, so Doctrine can't be used anymore on volumes smaller than 10 GB or so.

Ocramius:
> Is this really not going to be fixed until 1.5.0?

Yes, won't merge directory structure changes into a patch release anyway.

Ocramius:
I re-read through this multiple times and it looks :+1: 

Thanks for the effort put in this, @Tobion!

Tobion:
@Ocramius Thanks for merging. When do you plan to create a release tag?

Ocramius:
@Tobion I'm currently going through release notes

PR #95, opened 2015-10-08
Fix delete() and flushAll()

Tobion:
1. fix delete() return value when file does not exist: it must return true even if the cache entry doesn't exist. False should only be returned in case of a failed delete. Since the outcome of deleting a non-existing entry is what we wanted anyway (the entry doesn't exist anymore), it should return true. But the filesystem cache returned false because `unlink` returns false if the file does not exist. This was also broken for some other cache drivers.
2. fix flushAll() and getStats() for file cache to ignore dot files and clean up balancing directories: before it also iterated over dot files `.` and `..` (if the extension is empty) trying to delete them, which of course failed, but was silenced. and intermediate directories for balancing were never removed. flushAll only removed the files but not the directories leaving alot of garbage behind.
3. tests were added for the above things and tests for the existing behavior that caches in a shared directory but different extensions do not influence each other.
4. fix Memcached handling of `false` as data

### DIFF
--- a/tests/Doctrine/Tests/Common/Cache/BaseFileCacheTest.php
+++ b/tests/Doctrine/Tests/Common/Cache/BaseFileCacheTest.php
@@ -31,6 +31,8 @@ abstract class BaseFileCacheTest extends CacheTest
                 @rmdir($file->getRealPath());
             }
         }
+
+        @rmdir($this->directory);
     }
 
     protected function isSharedStorage()
--- a/tests/Doctrine/Tests/Common/Cache/BaseFileCacheTest.php
+++ b/tests/Doctrine/Tests/Common/Cache/BaseFileCacheTest.php
@@ -9,14 +9,27 @@ abstract class BaseFileCacheTest extends CacheTest
 {
     protected $directory;
 
-    public function setUp()
+    public function testFlushAllRemovesBalancingDirectories()
+    {
+        $cache = $this->_getCacheDriver();
+
+        $this->assertTrue($cache->save('key1', 1));
+        $this->assertTrue($cache->save('key2', 2));
+        $this->assertTrue($cache->flushAll());
+
+        $iterator = new \RecursiveIteratorIterator(new \RecursiveDirectoryIterator($this->directory, \FilesystemIterator::SKIP_DOTS), \RecursiveIteratorIterator::CHILD_FIRST);
+
+        $this->assertCount(0, $iterator);
+    }
+
+    protected function setUp()
     {
         do {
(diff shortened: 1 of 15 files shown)


tests/Doctrine/Tests/Common/Cache/BaseFileCacheTest.php
```diff
@@ -31,6 +44,8 @@ public function tearDown()
                 @rmdir($file->getRealPath());
             }
         }
+
+        @rmdir($this->directory);
```
Tobion: the cleanup was missing to remove the cache root dir thus every test run would create a new temp dir without deleting it.

Tobion:
Some cache providers fail the `testDeleteIsSuccessfulWhenKeyDoesNotExist` but others succeed like the ArrayCache. So there are some incosistencies in the behavior.

tests/Doctrine/Tests/Common/Cache/PhpFileCacheTest.php
```diff
@@ -10,6 +10,24 @@
...
+     * @dataProvider falseCastedValuesProvider
+    */
+    public function testFalseCastedValues($value)
+    {
+        if (0.0 === $value) {
+            $cache = $this->_getCacheDriver();
+
+            $this->assertTrue($cache->save('key', $value));
+            $this->assertTrue($cache->contains('key'));
+            $this->assertSame(0, $cache->fetch('key'), 'var_export exports float(0) as int(0) so we assert against 0 as integer');
```
Tobion: https://twitter.com/TobiasSchultze/status/652233645873360896

Tobion:
@Ocramius I fixed all drivers that were failing for deleting non-existing entries. The cache suit passes now.

tests/Doctrine/Tests/Common/Cache/CacheTest.php
```diff
@@ -8,25 +8,44 @@
 abstract class CacheTest extends \Doctrine\Tests\DoctrineTestCase
 {
     /**
-     * @dataProvider provideCrudValues
+     * @dataProvider provideDataToCache
      */
-    public function testBasicCrudOperations($value)
```
Tobion: this test ignored the data provider as $value has not been used at all.

tests/Doctrine/Tests/Common/Cache/PhpFileCacheTest.php
Ocramius: This is indeed painful :|

Ocramius:
Merged and backported into `1.4.x` via doctrine/cache@cdeb14a6feb34821a6f5fe05435ef04a68a7f147, thanks!

PR #96, opened 2015-10-09
[WIP] Namespaced cache decorator

Tobion:
The idea is to implement namespaces using a decorator. This makes much more sense than having namespaces built-into the generic CacheProvider because namspaces only make sense for shared caches. So non-shared caches like ArrayCache should not have namespace logic at all.

Also the need to add namespaces depends on the usage. If a developer uses a global cache like Redis but does not reuse it it many unrelated places (where he cannot ensure cache ids do not collide), there is no point in adding namespaces. This also has the benefit to not add the overhead when not needed. The current overhead comes from reading the namespace version on each fetch. So currently each fetch does two accesses. The version is cached in the CacheProvider to boost the performance, but that can obivously result in race conditions. So this is removed in the decorator.

Decorating a cache with this namespace logic will also make sure that deleting all cache entries using `flushAll()` will only remove the cache entries of the given namespace and not everything. This way one application/library can delete all of its cache without interfering with the shared cache of another app/library. This makes `ClearableCache::deleteAll` obsolete. Classes that use a cache should not have to care about `deleteAll` vs `flushAll`. They just want to delete the stuff. It's the task of the developer to make use of the namespace decorator when they have a global cache and different places where deletion can occur.

Todo
- [ ] Throw exception if decorated cache doesn't implement MultiGetCache or FlushableCache
- [ ] Add tests for decorator
- [ ] Deprecate `CacheProvider::setNamespace` and `CacheProvider::getNamespace`. In the next major version we can then remove namespacing from there and thus the cache providers that don't need it like ArrayCache will not by default inherit the namespace overhead.
- [ ] Update https://github.com/doctrine/DoctrineCacheBundle to use NamespacedCacheDecorator when a `namespace` config is used
- [ ] Update https://github.com/doctrine/doctrine2 to only use `flushAll` (or `deleteAll` if we want to settle on this name). This is mainly used in cache clear commands.

### DIFF
--- a/lib/Doctrine/Common/Cache/ClearableCache.php
+++ b/lib/Doctrine/Common/Cache/ClearableCache.php
@@ -28,6 +28,8 @@ namespace Doctrine\Common\Cache;
  * @link   www.doctrine-project.org
  * @since  1.4
  * @author Adirelle <adirelle@gmail.com>
+ *
+ * @deprecated Use NamespacedCacheDecorator instead which makes sure that flushAll() only removes cache entries in the current namespace
  */
 interface ClearableCache
 {
--- a/lib/Doctrine/Common/Cache/NamespacedCacheDecorator.php
+++ b/lib/Doctrine/Common/Cache/NamespacedCacheDecorator.php
@@ -1,0 +1,185 @@
+<?php
+/*
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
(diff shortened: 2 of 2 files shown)


andrerom:
Nice to see this being moved out.
Any chance drivers will be refactored to only implement the interfaces they support as part of this?  *as opposed to everything via an abstract*

andrerom:
Strike that, for bc cache provider will need to implement all anyway, so besides moving names spacing out it can stay as is.

Tobion:
I had another idea which would also make the decorator obsolete.

CacheProvider can accept a `CacheIdNamingStrategy` interface. We can then provide different implementations of it, e.g. `HashCacheIdNamingStrategy`, `TransparentCacheIdNamingStrategy` (not modifying the cache key) and `NamespacedCacheIdNamingStrategy`.

This would solve another problem at the same time: Currently the allowed cache ids are based on the used adapter. E.g. Memecached does not support whitespaces whereas others caches do. So if people would have to deal with special chars, they can just inject a hash strategy and are set whatever cache adapter they will use.

andrerom:
Time to move this forward? With symfony cache using this now it would be good to be able to avoid this overhead one way or another *(for me it is ;))*. personally I prefer decorator approach as it then doesn't add any cost when turned off at all.

And special rules on cache id's either needs to restrict interface of all, or be normalized in specific adapter imo. Forcing user to match correct name strategy service with adapter is asking for mistakes to happen.

yellow1912:
Any hope for this to be rolled out anytime soon? I need to save to redis in my application but then retrieve from other place. (lua code). The versioning is messing up the expected key.

yellow1912:
@Tobion I think your idea with CacheIdNamingStrategy can work really well. We can then easily support the configuration in DoctrineCacheBundle as well, by specifying the naming service we may want to use.

Tobion:
Closing as I don't have intention to finish this one.

PR #97, opened 2015-10-10
unify bool(ean) and add multi get support for wincache

Tobion:


lib/Doctrine/Common/Cache/ApcCache.php
```diff
@@ -53,7 +53,7 @@ protected function doContains($id)
      */
     protected function doSave($id, $data, $lifeTime = 0)
     {
-        return (bool) apc_store($id, $data, (int) $lifeTime);
+        return apc_store($id, $data, $lifeTime);
```
Tobion: it returns bool anyway with a single id

lib/Doctrine/Common/Cache/WinCacheCache.php
```diff
@@ -75,6 +75,14 @@ protected function doFlush()
     /**
      * {@inheritdoc}
      */
+    protected function doFetchMultiple(array $keys)
+    {
+        return wincache_ucache_get($keys);
```
Tobion: http://php.net/manual/en/function.wincache-ucache-get.php#refsect1-function.wincache-ucache-get-returnvalues

Ocramius:
Rebased and manually merged via doctrine/cache@16e39bfd0921ba4138cb97344c4fdd54c185bf6f, thanks!

PR #98, opened 2015-10-10
add travis cache and fix apcu installation

Tobion:
- use travis cache
- fix apcu loading for php 5.5 and 5.6 which was broken and thus did not test ApcCache there
- use autoload-dev
- use phpunit `@requires` annotation


Ocramius:
Merged after manual rebase via doctrine/cache@3655ec928cd2cd7e7591e9f416bf402dfd228c62, thanks @Tobion!

PR #103, opened 2015-10-29
PredisCache - added getClient method

ovr:
Hey!

Sometimes I needed to get PredisClient from Cache
So I can use 2 service in $di but I think this will be usefull

### DIFF
--- a/lib/Doctrine/Common/Cache/PredisCache.php
+++ b/lib/Doctrine/Common/Cache/PredisCache.php
@@ -25,6 +25,14 @@ class PredisCache extends CacheProvider
     {
         $this->client = $client;
     }
+    
+    /**
+     * @return Client
+     */
+    public function getClient()
+    {
+        return $this->client;
+    }
 
     /**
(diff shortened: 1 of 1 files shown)


Ocramius:
@ovr while I know the other adapters implement similar methods, I really suggest not going this way further on.
Unless there is a very strong need for the getter (can't inject), this should be avoided.

Ocramius:
Indeed, after reading through it again, won't go down this way.

Closing

ovr:
@Ocramius 
Main problem, Doctrine cache doesnt provide many method of Predis client

For now I am using something likes this

```php
$di['predis'] = function ($di) {return Predis();}
$di['cache'] = function($di) {new Doctrine\Cache\PRedis($di['predis'])}
```

But I hate this way, Why are you not interesting to allow merge this method? This will be very usefull, because I am using cache service only for Doctrine....

Ocramius:
@ovr how you are doing it there is the right way..

PR #104, opened 2015-10-30
fetchMultiple problems with null and falsey values

mikesul:
If a null is stored in the cache, fetchMultiple will not return it as if it does not exist in the cache. It is returned internally by doFetchMultiple, but then unexpectedly filtered out by the isset check.

This applies the same approach used in ArrayCache to avoid calling array_key_exists unnecessarily.

### DIFF
--- a/lib/Doctrine/Common/Cache/PredisCache.php
+++ b/lib/Doctrine/Common/Cache/PredisCache.php
@@ -45,17 +45,8 @@ class PredisCache extends CacheProvider
     protected function doFetchMultiple(array $keys)
     {
         $fetchedItems = call_user_func_array(array($this->client, 'mget'), $keys);
-        $foundItems = array();
-        $resultCounter = 0;
-
-        foreach ($keys AS $key) {
-            if ($fetchedItems[$resultCounter] !== null) {
-                $foundItems[$key] = unserialize($fetchedItems[$resultCounter]);
-            }
-            $resultCounter++;
-        }
-
-        return $foundItems;
+        
+        return array_map('unserialize', array_filter(array_combine($keys, $fetchedItems)));
     }
 
     /**
--- a/lib/Doctrine/Common/Cache/PredisCache.php
+++ b/lib/Doctrine/Common/Cache/PredisCache.php
@@ -45,7 +45,7 @@ class PredisCache extends CacheProvider
     protected function doFetchMultiple(array $keys)
     {
         $fetchedItems = call_user_func_array(array($this->client, 'mget'), $keys);
-        
+
         return array_map('unserialize', array_filter(array_combine($keys, $fetchedItems)));
     }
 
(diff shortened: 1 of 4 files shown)


Ocramius:
@mikesul please check the failures reported by travis-ci

mikesul:
The 2 failures appear to be the Predis and Redis backends. I'm not familiar with the specific details of those backends and I don't have a local setup of them to test particularly well.

However, both of them have implementations of doFetchMultiple which calls array_filter with no callback, so it will remove all falsey values. That looks like a separate bug (as it will strip out 0, empty string, empty array, etc) but I'd need to get a setup to do more testing with them.

mikesul:
I've spun the fetchMulti null test into a generic test for falsey values. With my previous commit in place, all providers passed except for Redis and Predis. These have both been updated to handle falsey values as expected and the build passes.

lib/Doctrine/Common/Cache/PredisCache.php
```diff
@@ -45,8 +45,17 @@ protected function doFetch($id)
...
+        $resultCounter = 0;
+
+        foreach ($keys AS $key) {
+            if ($fetchedItems[$resultCounter] !== null) {
+                $foundItems[$key] = unserialize($fetchedItems[$resultCounter]);
+            }
+            $resultCounter++;
+        }
 
-        return array_filter(array_combine($keys, array_map('unserialize', $fetchedItems)));
```
Ocramius: You could actually `array_filter()` before passing to `unserialize()` instead of re-writing the whole body as it was above

lib/Doctrine/Common/Cache/RedisCache.php
```diff
@@ -71,7 +71,18 @@ protected function doFetch($id)
      */
     protected function doFetchMultiple(array $keys)
     {
-        return array_filter(array_combine($keys, $this->redis->mget($keys)));
```
Ocramius: Same suggestion as above

Ocramius:
Looks good, except for semantic change requests. I can apply them for you on merge, if you prefer

mikesul:
I've gone back to the array_combine/array_filter approach as you suggested. The Predis approach is a little simpler because we can do a basic filter pre-unserialize. Since Redis is outputting the actual types, we need to add an explicit !== false check.

Found a couple code standards faults in my commits and sorted them as well. Hopefully this should all be up to scratch for merging now.

Ocramius:
@mikesul thanks, merged into:

`1.4.4`: doctrine/cache@e30f23a90993f55cdc4817e5f5d67d73d7c9161e
`1.5.1`: doctrine/cache@91b06ecf58c7723a177e3957e5b1b3d532ccc9cb
`1.6.0` (`master`) : doctrine/cache@367a071758e771830439a18d312ac59490ef3cf5

tests/Doctrine/Tests/Common/Cache/CacheTest.php
```diff
@@ -96,6 +96,31 @@ public function testFetchMultiWithEmptyKeysArray()
         );
     }
 
+    public function testFetchMultiWithFalsey()
+    {
+        $cache = $this->_getCacheDriver();
+
+        $cache->deleteAll();
+
+        $values = array(
```
Tobion: This should have used `provideDataToCache` instead which could be reused and also covers more.

PR #105, opened 2015-11-18
Fix fetch multiple with false

Tobion:
When fixing https://github.com/doctrine/cache/pull/104/files#r43902042 I found that the `fetchMultiple` is broken, when the cached data is `false` which I fixed as well.

### DIFF
--- a/tests/Doctrine/Tests/Common/Cache/CacheTest.php
+++ b/tests/Doctrine/Tests/Common/Cache/CacheTest.php
@@ -119,6 +119,8 @@ abstract class CacheTest extends \Doctrine\Tests\DoctrineTestCase
         return array(
             'array' => array(array('one', 2, 3.01)),
             'string' => array('value'),
+            'string_invalid_utf8' => array("\xc3\x28"),
+            'string_null_byte' => array('with'."\0".'null char'),
             'integer' => array(1),
             'float' => array(1.5),
             'object' => array(new ArrayObject(array('one', 2, 3.01))),
@@ -131,7 +133,7 @@ abstract class CacheTest extends \Doctrine\Tests\DoctrineTestCase
             'string_zero' => array('0'),
             'integer_zero' => array(0),
             'float_zero' => array(0.0),
-            'string_empty' => array('')
+            'string_empty' => array(''),
         );
     }
 
@@ -139,6 +141,7 @@ abstract class CacheTest extends \Doctrine\Tests\DoctrineTestCase
     {
         $cache = $this->_getCacheDriver();
 
+        $cache->delete('key');
         $this->assertFalse($cache->contains('key'));
         $this->assertTrue($cache->delete('key'));
     }
--- a/tests/Doctrine/Tests/Common/Cache/CacheTest.php
+++ b/tests/Doctrine/Tests/Common/Cache/CacheTest.php
@@ -233,6 +233,30 @@ abstract class CacheTest extends \Doctrine\Tests\DoctrineTestCase
         );
     }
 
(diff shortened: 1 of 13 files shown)


Ocramius:
@Tobion travis seems to disagree with the falsey-values changes

Tobion:
@Ocramius fixed and refactored some more tests. PHP 7 test has a segfault but that should not be related to this PR.

tests/Doctrine/Tests/Common/Cache/CacheTest.php
```diff
@@ -63,72 +63,68 @@ public function testCacheKeyIsCaseSensitive()
...
 
-    public function testFetchMulti()
+    public function testFetchMultiple()
     {
         $cache = $this->_getCacheDriver();
 
-        $cache->deleteAll();
+        $values = $this->provideDataToCache();
+        foreach ($values as $key => $value) {
+            $cache->save($key, $value[0]);
```
Ocramius: Just reuse `$values[$key]` from the previous assignment?

tests/Doctrine/Tests/Common/Cache/CacheTest.php
```diff
@@ -236,6 +233,30 @@ public function provideCacheIds()
         );
     }
 
+    public function testLifetime()
+    {
+        $cache = $this->_getCacheDriver();
+        $cache->save('expire', 'value', 1);
+        $this->assertTrue($cache->contains('expire'), 'Data should not be expired yet');
+        sleep(2);
```
Ocramius: Not acceptable IMO

tests/Doctrine/Tests/Common/Cache/CacheTest.php
```diff
@@ -236,6 +233,30 @@ public function provideCacheIds()
...
+        $this->assertFalse($cache->contains('expire'), 'Data should be expired');
+    }
+
+    public function testNoExpire()
+    {
+        $cache = $this->_getCacheDriver();
+        $cache->save('noexpire', 'value', 0);
+        sleep(1);
```
Ocramius: Same as above

tests/Doctrine/Tests/Common/Cache/CacheTest.php
Ocramius: Note: aware they were in the test suite before (they shouldn't have been tho)

tests/Doctrine/Tests/Common/Cache/CacheTest.php
Tobion: I don't understand what you mean.

tests/Doctrine/Tests/Common/Cache/CacheTest.php
Tobion: What alternative do you have in mind to test lifetime?

Tobion: As you can see the test already reveaed two inconsistencies in the cache implementations.

tests/Doctrine/Tests/Common/Cache/CacheTest.php
Ocramius: Sorry, comment was on the wrong line.

tests/Doctrine/Tests/Common/Cache/CacheTest.php
Ocramius: Checked again - this can be accepted for now, but I will add a `@todo` line to it.

Ocramius:
@Tobion thanks!

PR #106, opened 2015-11-18
Fix. Invalid filename length in FileCache:getFilename()

fgrandjean:
Only the $id length was checked for not exceeding the 255 chars limitations.
Using the sha256 hash without taking into account the "_" prefix and the custom extension could lead to exceed that limit.


Ocramius:
@fgrandjean was this supposed to be closed?

fgrandjean:
Yes. Sorry, I though i had fixed the issue but not yet.... I will reopen when the fix will be working properly.

PR #107, opened 2015-11-21
fix file cache naming under windows surpassing MAX_PATH

Tobion:
I realised that doctrine annotation cache (which uses long cache keys) is failing in symfony under windows.
Its because in Windows the whole path is limited to 260 chars. See http://stackoverflow.com/questions/1880321/why-does-the-260-character-path-length-limit-exist-in-windows

Also fixed travis tests with apcu due to new releases of apcu for php 5 and 7.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,8 @@ services:
 before_install:
     - if [[ $TRAVIS_PHP_VERSION != 'hhvm' && $TRAVIS_PHP_VERSION != '7.0' ]]; then pecl install riak-beta; fi;
     - if [[ $TRAVIS_PHP_VERSION =~ 5.[34] ]] ; then echo 'extension="apc.so"' >> ./tests/travis/php.ini; fi;
-    - if [[ $TRAVIS_PHP_VERSION =~ 5.[56] ]] ; then pecl config-set preferred_state beta; printf "yes\n" | pecl install apcu; fi;
+    - if [[ $TRAVIS_PHP_VERSION =~ 5.[56] ]] ; then echo yes | pecl install apcu; fi;
+    - if [[ $TRAVIS_PHP_VERSION = 7.* ]] ; then pecl config-set preferred_state beta; echo yes | pecl install apcu; fi;
     - if [[ $TRAVIS_PHP_VERSION != 'hhvm' ]]; then phpenv config-add ./tests/travis/php.ini; fi;
 
 install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,14 +25,14 @@ before_install:
     - if [[ $TRAVIS_PHP_VERSION != 'hhvm' && $TRAVIS_PHP_VERSION != '7.0' ]]; then pecl install riak-beta; fi;
     - if [[ $TRAVIS_PHP_VERSION =~ 5.[34] ]] ; then echo 'extension="apc.so"' >> ./tests/travis/php.ini; fi;
     - if [[ $TRAVIS_PHP_VERSION =~ 5.[56] ]] ; then echo yes | pecl install apcu; fi;
-    - if [[ $TRAVIS_PHP_VERSION = 7.* ]] ; then phpenv config-rm xdebug.ini; pecl config-set preferred_state beta; echo yes | pecl install apcu; fi;
+    - if [[ $TRAVIS_PHP_VERSION = 7.* ]] ; then pecl config-set preferred_state beta; echo yes | pecl install apcu; fi;
     - if [[ $TRAVIS_PHP_VERSION != 'hhvm' ]]; then phpenv config-add ./tests/travis/php.ini; fi;
 
 install:
     - travis_retry composer install
 
 script:
-    - ./vendor/bin/phpunit -c ./tests/travis/phpunit.travis.xml -v --debug
+    - ./vendor/bin/phpunit -c ./tests/travis/phpunit.travis.xml -v
 
 after_script:
     - php vendor/bin/coveralls -v
@@ -41,3 +41,4 @@ matrix:
     fast_finish: true
     allow_failures:
         - php: hhvm
+        - php: 7.0
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,14 +25,14 @@ before_install:
     - if [[ $TRAVIS_PHP_VERSION != 'hhvm' && $TRAVIS_PHP_VERSION != '7.0' ]]; then pecl install riak-beta; fi;
     - if [[ $TRAVIS_PHP_VERSION =~ 5.[34] ]] ; then echo 'extension="apc.so"' >> ./tests/travis/php.ini; fi;
     - if [[ $TRAVIS_PHP_VERSION =~ 5.[56] ]] ; then echo yes | pecl install apcu; fi;
-    - if [[ $TRAVIS_PHP_VERSION = 7.* ]] ; then pecl config-set preferred_state beta; echo yes | pecl install apcu; fi;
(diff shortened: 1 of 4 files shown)


Tobion:
I put php 7 to allowed failures as long as the seg fault happens. I thought it might be xdebug but the error happens also without xdebug:
- with xdebug:

> /home/travis/build.sh: line 45:  8445 Segmentation fault      (core dumped) ./vendor/bin/phpunit -c ./tests/travis/phpunit.travis.xml -v

- without xdebug: 

> Starting test 'Doctrine\Tests\Common\Cache\VoidCacheTest::testShouldAlwaysReturnFalseOnContains'.
/home/travis/build.sh: line 45:  8474 Bus error               (core dumped) ./vendor/bin/phpunit -c ./tests/travis/phpunit.travis.xml -v --debug

Tobion:
@Ocramius would be nice to merge this before the symfony release and to create a tag

MisterGrouik:
Has it been merged ? Still have the problem with Symfony 2.8

Ocramius:
Not yet in a release.

Ocramius:
@Tobion merged, thanks!

TracKer:
Seems like the problem is still there... :(

Ocramius:
@TracKer more details?

TracKer:
Here is the path that I have:

D:\Work\OpenServer\domains\atest.local\var\cache\dev\annotations\c2\5b4a6176696572456775696c757a5c42756e646c655c4561737941646d696e42756e646c655c436f6e74726f6c6c65725c41646d696e436f6e74726f6c6c65722366696e64416c6c405b416e6e6f745d5d5b315d.doctrinecache.data

It's 256 chars long, but I am getting an error while clearing cache, like described in #16783.

Tobion:
@TracKer you need https://github.com/symfony/symfony/pull/16829 as well which is not released yet.

TracKer:
Thanks @Tobion, just checked by switching symfony to dev version - cache is clearing as expected!
:+1:

gnat42:
This unfortunately doesn't seem to fully fix the problem. So for example I have a system deployed on windows and was hit by this bug for two reasons. First the filename + directory path were longer than the minimum 260 chars. This was partially because of symfony renaming the cache directory to test_old (adding 4 chars). However after applying their fix and this one, I could clear the cache but not run application availability tests (via phpunit and symfony's "browser"). It would fail on

https://github.com/doctrine/cache/blob/master/lib/Doctrine/Common/Cache/FilesystemCache.php#L53

where it ran fopen($filename,"r"); php threw a Invalid Argument error/exception. I wrote a *super simple* test.
```
$filename = 'C:\Users\nuviadmin\nuvi\app\cache\test\annotations\d3\5b53796d666f6e795c42756e646c655c4672616d65776f726b42756e646c655c436f6e74726f6c6c65725c436f6e74726f6c6c65722363726561746541636365737344656e696564457863657074696f6e405b416e6e6f745d5d5b315d.doctrinecache.data';

echo is_file($filename) ? "Is File\n":"Is NOT File\n";
echo "len: ".strlen($filename)."\n";
$resource = fopen($filename,'r');

if(!$resource) {
  echo "Not opened\n".print_r(error_get_last(),true);
} else { 
  echo "Opened\n";
}
```

This would result in is_file returning true, but fopen not opening the file with a Invalid Argument.

I went back to lib/Doctrine/Common/Cache/FileCache.php and changed 259 to 255. Everything works after that. Why I have NO idea. However that 259 needs to be dropped down at least by one to 258 like mentioned in #113 

It doesn't fully work otherwise. Please drop it to 258.

Tobion:
Please open a bug report on php as well.

gnat42:
There already is a bug filed. https://bugs.php.net/bug.php?id=70943

Tobion:
I added a comment on the PHP bug:


I confirm the problem and investigated which filesystem functions work with different lengths of filenames. In theory filesnames up to 259 length should work on Windows as it is below the 260 MAX_PATH length (including terminating null char). I could find out that some functions work with length 259 (touch, rename, file_exists etc), while other functions fail (copy, fopen, realpath, file_put_contents) for the same filename. So for these functions PHP does seem to do something inconsistently in a bad way which causes hard to debug problems..


Test script:
------------
```php
<?php

$tmpDir = sys_get_temp_dir() . DIRECTORY_SEPARATOR;
$tmpDirLen = strlen($tmpDir);

foreach (array(258, 259, 260) as $fileLen) {
    $filename = $tmpDir . str_repeat('a', $fileLen - $tmpDirLen);
    var_dump($filename);

    echo 'touch: ';
    var_dump(@touch($filename));

    echo 'chmod: ';
    var_dump(@chmod($filename, 0666));

    echo 'file_exists: ';
    var_dump(file_exists($filename));

    echo 'is_file: ';
    var_dump(is_file($filename));

    echo 'stat: ';
    var_dump(@stat($filename) ? true : false);

    echo 'realpath: ';
    var_dump(realpath($filename) ? true : false);

    echo 'fopen: ';
    var_dump(($handle = @fopen($filename, 'r')) ? true : false);

    if ($handle) {
        fclose($handle);
    }

    echo 'unlink: ';
    var_dump(@unlink($filename));

    $tmpFile = tempnam(sys_get_temp_dir(), 'tmp');

    echo 'copy: ';
    var_dump(@copy($tmpFile, $filename));

    echo 'file_put_contents: ';
    var_dump(@file_put_contents($filename, 'test data') ? true : false);

    echo 'rename: ';
    var_dump($rename = @rename($tmpFile, $filename));

    unlink($rename ? $filename : $tmpFile);

    echo "\n\n";
}
```

Expected result:
----------------
```
string(258) "..."
touch: bool(true)
chmod: bool(true)
file_exists: bool(true)
is_file: bool(true)
stat: bool(true)
realpath: bool(true)
fopen: bool(true)
unlink: bool(true)
copy: bool(true)
file_put_contents: bool(true)
rename: bool(true)


string(259) "..."
touch: bool(true)
chmod: bool(true)
file_exists: bool(true)
is_file: bool(true)
stat: bool(true)
realpath: bool(true)
fopen: bool(true)
unlink: bool(true)
copy: bool(true)
file_put_contents: bool(true)
rename: bool(true)


string(260) "..."
touch: bool(false)
chmod: bool(false)
file_exists: bool(false)
is_file: bool(false)
stat: bool(false)
realpath: bool(false)
fopen: bool(false)
unlink: bool(false)
copy: bool(false)
file_put_contents: bool(false)
rename: bool(false)
```

Actual result:
--------------
```
string(258) "..."
touch: bool(true)
chmod: bool(true)
file_exists: bool(true)
is_file: bool(true)
stat: bool(true)
realpath: bool(true)
fopen: bool(true)
unlink: bool(true)
copy: bool(true)
file_put_contents: bool(true)
rename: bool(true)


string(259) "..."
touch: bool(true)
chmod: bool(true)
file_exists: bool(true)
is_file: bool(true)
stat: bool(true)
realpath: bool(false)
fopen: bool(false)
unlink: bool(true)
copy: bool(false)
file_put_contents: bool(false)
rename: bool(true)


string(260) "..."
touch: bool(false)
chmod: bool(false)
file_exists: bool(false)
is_file: bool(false)
stat: bool(false)
realpath: bool(false)
fopen: bool(false)
unlink: bool(false)
copy: bool(false)
file_put_contents: bool(false)
rename: bool(false)
```

emudojo:
This is still happening windows 10x64 and PHP 7, cache files are too large.

pculka:
can only confirm Windows 10x64, PHP7, cache filenames are too long and failing. This is annoying a bit...

gnat42:
It seems they fixed it in 7.1 according the PHP bug but doesn't look like it'll be back ported. So something needs to be fixed in this code to handle the issue... then both bugs are closed....

pculka:
I just tried 7.1RC1 on windows, doesn't really work ... issue still remains

PR #108, opened 2015-11-25
Avoid MongoCursorException with MongoCache

remialvado:
Without this commit, we sometimes have error like ```E11000 duplicate key error collection: cache.cache index: ​_id_​ dup key: { ...``` I don't know exactly what happen inside MongoDB to throw this error but the Cache interface is pretty clear that it should not throw any Exception but return a boolean to express that the cache has been saved or not


Ocramius:
@remialvado I added tests and applied some CS fixes: thanks!

remialvado:
Thanks !! Sorry for the missing test :(

PR #109, opened 2015-12-03
Cleanup: drop unsupported php versions

Ocramius:
Because http://php.net/supported-versions.php

Also: tempted to drop 5.5 as well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,8 +8,6 @@ cache:
         - $HOME/.composer/cache
 
 php:
-    - 5.3
-    - 5.4
     - 5.5
     - 5.6
     - 7.0
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ cache:
         - $HOME/.composer/cache
 
 php:
+    - 5.5
     - 5.6
     - 7.0
     - hhvm
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,6 @@ cache:
         - $HOME/.composer/cache
 
 php:
-    - 5.5
     - 5.6
     - 7.0
     - hhvm
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,3 +39,4 @@ matrix:
     fast_finish: true
     allow_failures:
         - php: hhvm
+        - php: 7.0
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,4 +40,3 @@ matrix:
     fast_finish: true
     allow_failures:
         - php: hhvm
-        - php: 7.0
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,6 @@ services:
 before_install:
     - if [[ $TRAVIS_PHP_VERSION != 'hhvm' ]] ; then pecl channel-update pecl.php.net; fi;
     - if [[ $TRAVIS_PHP_VERSION != 'hhvm' && $TRAVIS_PHP_VERSION != '7.0' ]]; then pecl install riak-beta; fi;
-    - if [[ $TRAVIS_PHP_VERSION =~ 5.[34] ]] ; then echo 'extension="apc.so"' >> ./tests/travis/php.ini; fi;
     - if [[ $TRAVIS_PHP_VERSION =~ 5.[56] ]] ; then echo yes | pecl install apcu-4.0.10; fi;
     - if [[ $TRAVIS_PHP_VERSION = 7.* ]] ; then pecl config-set preferred_state beta; echo yes | pecl install apcu; fi;
     - if [[ $TRAVIS_PHP_VERSION != 'hhvm' ]]; then phpenv config-add ./tests/travis/php.ini; fi;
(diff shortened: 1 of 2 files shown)


guilhermeblanco:
-1 on dropping PHP 5.5.

Major OS distros support LTS and we should be inline with their versions too

.travis.yml
```diff
@@ -8,9 +8,6 @@ cache:
         - $HOME/.composer/cache
 
 php:
-    - 5.3
-    - 5.4
-    - 5.5
```
guilhermeblanco: This does more than your ticket claims to do (you actually dropped PHP 5.5 support

composer.json
```diff
@@ -13,10 +13,10 @@
         {"name": "Johannes Schmitt", "email": "schmittjoh@gmail.com"}
     ],
     "require": {
-        "php": ">=5.3.2"
+        "php": "~5.6|~7.0"
```
guilhermeblanco: `~5.5|~7.0`

composer.json
```diff
@@ -13,10 +13,10 @@
...
     },
     "require-dev": {
-        "phpunit/phpunit":         ">=3.7",
+        "phpunit/phpunit":         "~4.0|~5.0",
```
guilhermeblanco: `~4.8|~5.0`

Ocramius:
Will restore 5.5 for 1.6: absolutely gonna kill it in further minor versions though.

Ocramius:
@guilhermeblanco restored

.travis.yml
```diff
@@ -8,8 +8,6 @@ cache:
         - $HOME/.composer/cache
 
 php:
-    - 5.3
-    - 5.4
```
Tobion: line `if [[ $TRAVIS_PHP_VERSION =~ 5.[34] ]]` can be removed as well

.travis.yml
```diff
@@ -41,4 +39,3 @@ matrix:
     fast_finish: true
     allow_failures:
         - php: hhvm
-        - php: 7.0
```
Tobion: php7 still seg fault

stof:
@Ocramius Symfony 3 minimum version is 5.5.9, and it depends on doctrine/cache. So you will either have to keep support for PHP 5.5 in following doctrine/cache versions, or provide LTS support for 1.6 for the next few years (so that Symfony can ship with a maintained version of doctrine/cache).
I'm not sure supporting PHP 5.5 actually requires much more work than supporting 5.6 (are you planning to use any 5.6+ feature in the library for something harder to do on 5.5 ?)

stof:
And you should check what broke the mongo tests

dunglas:
I personally don't care but even if the EOL of PHP 5.4 has been reached, it is still supported in Debian Wheezy until may 2018: https://packages.debian.org/Wheezy/php5

Ocramius:
We're not a support shop, folks! We do security support and we support what
we can, but we can't nor will commit to having LTS stuff anyway. Symfony
LTS can simply rely on 1.6, and we'll fix security or very severe issues if
they pop up.

Staying in sync with the PHP releases (official ones) reduces our surface
of engagement, which is a better commitment than just stating that we will
support more (we simply don't have the resources for that). In general,
expect less supported versions by all packages in future, while we switch
non-current versions to security-only.

GeeH:
You should totally drop support for all PHP 5.x versions.

aequasi:
-1 on removing 5.5. It's not EOLed yet. Try again in 6 months

ThaDafinser:
I love this drop of 5.5 \o/

Ocramius:
To those still commenting about 5.5, it was reverted at doctrine/cache@9ecf9239b0bce8b39637c47aa8115ae9bb443d8c last week.

.travis.yml
Ocramius: done

.travis.yml
Ocramius: Currently broken as of travis-ci/travis-ci#5256 and travis-ci/travis-ci#5207

GrahamCampbell:
PHP 5.5 goes EOL in about 6 months. Maybe just wait till then?

GrahamCampbell:
Personally, I'm not massively bothered, as long as php 5.6 support remains for at least 1 more year.

Ocramius:
@GrahamCampbell 5.5 is still supported in this PR

GrahamCampbell:
Yeh, I know. ;) I was just referring to your comment when you said you'd like to drop it in the very near future,

GrahamCampbell:
I'm very glad php 5.3 support is finally dropped. ;)

Ocramius:
@GrahamCampbell we will likely move to 7.0 for 1.7.x, since 5.6 is not used in any linux LTS (as far as I know)

GrahamCampbell:
That's a very big jump. ;(

Ocramius:
@GrahamCampbell not a very big one, to be honest. See https://github.com/doctrine/cache/pull/109#issuecomment-162181547 againagain

dunglas:
@Ocramius Just for info Debian Jessie (LTS - end of life in 2020) is shipped with PHP 5.6: https://packages.debian.org/jessie/php5

Ocramius:
@dunglas ouch :-\

GrahamCampbell:
We'll be on php 9 by then, lol.

PR #110, opened 2015-12-03
[LifeTime] Allows to set a maxLifeTime in cacheProvider

peelandsee:
Allows to set a maximum life-time at cacheProvider level.

Useful to set a service-wide limit,
not denying users to define their own (lower) limit for each save() call.

### DIFF
--- a/lib/Doctrine/Common/Cache/CacheProvider.php
+++ b/lib/Doctrine/Common/Cache/CacheProvider.php
@@ -48,6 +48,11 @@ abstract class CacheProvider implements Cache, FlushableCache, ClearableCache, M
     private $namespaceVersion;
 
     /**
+     * @var integer|bool
+     */
+    private $maxLifeTime = false;
+
+    /**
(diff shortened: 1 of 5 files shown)


Ocramius:
@peelandsee even if useful, this is quite an edge case.
Also, you should isolate it via a wrapper instead (all the logic can be implemented via a decorator):

```php
class MaxTtlLimitingCache implements Cache
{
    private $wrapped;
    private $maxTtl;
    public function __construct(Cache $wrapped, int $maxTtl)
    {
        $this->wrapped = $wrapped;
        $this->maxTtl = $maxTtl;
    }

    // ... 

    public function save($id, $data, $ttl = 0)
    {
        return $this->wrapped->save($id, $data, min($ttl, $this->maxTtl));
    }
}
```

Sorry, closing as `won't fix` here

peelandsee:
@Ocramius thanks for reviewing

PR #111, opened 2015-12-03
Cleanup: ext-apcu compat cleanup

Ocramius:
See krakjoe/apcu#42

I need somebody to review possible impact of this PR though, as conflicting doesn't necessarily mean that the users will use an `ApcCache`

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,8 @@
         "predis/predis":           "~1.0"
     },
     "conflict": {
-        "doctrine/common": ">2.2,<2.4"
+        "doctrine/common": ">2.2,<2.4",
+        "ext-apcu":        "<4.0.8"
     },
     "autoload": {
         "psr-4": { "Doctrine\\Common\\Cache\\": "lib/Doctrine/Common/Cache" }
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -21,7 +21,8 @@
         "predis/predis":           "~1.0"
     },
     "conflict": {
-        "doctrine/common": ">2.2,<2.4"
+        "doctrine/common": ">2.2,<2.4",
+        "ext-apcu":        "<4.0.8"
```
Ocramius: As per @beberlei's review, this may cause issues, I'm just not aware of the impact.

guilhermeblanco:
+1 from me

Ocramius:
As suggested by @beberlei, throwing an exception in `__construct()` when an incompatible `apcu` version is detected may be a better solution

composer.json
stof: this would forbid people to install this version if they have apcu installed, even though they want to use the RedisCache. So this has indeed an unwanted impact

GrahamCampbell: Maybe it's best to add this kind of check at runtime in the apc cache class rather than preventing installation of the entire package?

Ocramius:
Closing: I think #112 tackling this is a better approach.

PR #112, opened 2015-12-05
Native APCu support

dunglas:
Make the APC class supporting both old APC and new APCu signatures.

### DIFF
--- a/lib/Doctrine/Common/Cache/ApcCache.php
+++ b/lib/Doctrine/Common/Cache/ApcCache.php
@@ -20,7 +20,7 @@
 namespace Doctrine\Common\Cache;
 
 /**
- * APC cache provider.
+ * APC and APCu cache provider.
  *
  * @link   www.doctrine-project.org
  * @since  2.0
@@ -29,15 +29,29 @@ namespace Doctrine\Common\Cache;
  * @author Jonathan Wage <jonwage@gmail.com>
  * @author Roman Borschel <roman@code-factory.org>
  * @author David Abdemoulaie <dave@hobodave.com>
+ * @author Kévin Dunglas <dunglas@gmail.com>
  */
 class ApcCache extends CacheProvider
 {
+    /**
+     * @var bool
+     */
+    private $apcu;
+
+    /**
+     * @param bool $useApc Uses APC even if APCu is available
+     */
+    public function __construct($useApc = false)
+    {
+        $this->apcu = !$useApc && function_exists('apcu_fetch');
+    }
+
     /**
      * {@inheritdoc}
      */
     protected function doFetch($id)
     {
-        return apc_fetch($id);
+        return $this->apcu ? apcu_fetch($id) : apc_fetch($id);
     }
 
     /**
@@ -45,7 +59,7 @@ class ApcCache extends CacheProvider
      */
     protected function doContains($id)
     {
-        return apc_exists($id);
+        return $this->apcu ? apcu_exists($id) : apc_exists($id);
     }
 
     /**
@@ -53,7 +67,7 @@ class ApcCache extends CacheProvider
      */
     protected function doSave($id, $data, $lifeTime = 0)
     {
-        return apc_store($id, $data, $lifeTime);
+        return $this->apcu ? apcu_store($id, $data, $lifeTime) : apc_store($id, $data, $lifeTime);
     }
 
     /**
@@ -62,6 +76,10 @@ class ApcCache extends CacheProvider
     protected function doDelete($id)
     {
         // apc_delete returns false if the id does not exist
+        if ($this->apcu) {
+            return apcu_delete($id) || ! apcu_exists($id);
+        }
+
         return apc_delete($id) || ! apc_exists($id);
     }
 
@@ -70,6 +88,10 @@ class ApcCache extends CacheProvider
      */
     protected function doFlush()
     {
+        if ($this->apcu) {
+            return apcu_clear_cache();
+        }
+
         return apc_clear_cache() && apc_clear_cache('user');
     }
 
@@ -78,7 +100,7 @@ class ApcCache extends CacheProvider
      */
     protected function doFetchMultiple(array $keys)
     {
-        return apc_fetch($keys);
+        return $this->apcu ? apcu_fetch($keys) : apc_fetch($keys);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


layanto:
Personally prefer implementation similar to this, not using call_user_func

https://github.com/illuminate/cache/blob/master/ApcWrapper.php

Tobion:
Do all those methods behave exactly the same? And if so, why did they change the naming?

layanto:
APC has both opcache and userland cache. Since PHP 5.5+ includes opcache, APC opcache is no longer needed. APCu is APC with opcache removed, leaving just the userland cache; hence the same naming for functions except apc becomes apcu. For PHP 5.5 and 5.6, APCu includes APC backward compatibility feature, all apc_* functions internally call apcu_*. For PHP 7, backward compatibility mode is not enabled by default (and currently still not working for Windows).

layanto:
APC Polyfill
https://github.com/lavoiesl/apc-polyfill/blob/master/polyfill.php

lib/Doctrine/Common/Cache/ApcCache.php
```diff
@@ -86,8 +98,8 @@ protected function doFetchMultiple(array $keys)
      */
     protected function doGetStats()
     {
-        $info = apc_cache_info('', true);
-        $sma  = apc_sma_info();
+        $info = 'apcu' === $this->ext ? apc_cache_info(true) : apc_cache_info('', true);
```
stof: don't they have an ``apcu`` flavor of the function ?

stof:
I'm not a fan of the ``call_user_func`` usage either

dunglas:
@stof sorry I pushed a new commit in the meantime changing the implementation:

* Don't rely on `call_user_func` anymore
* Allow to explicitly use APC functions even if APCu functions are available using a constructor flag.

layanto:
What is the use case for explicit APC flag? I believe with apcu internally apc_* will call apcu_* anyway with bc extension.

dunglas:
@layanto for instance if the polyfill is installed, it allows to avoid unnecessary function calls. But it's maybe useless.

Tobion:
Imo the implementation should be different (which avoids the above problem):
- As support for php < 5.5 is going to be dropped (#109), `ApcCache` should be deprecated instead of adding features to it
- We provide a new `ApcuCache` just with the new methods

dunglas:
@Tobion doing so will require to update the code base (instead of just running `composer update`) of almost all projects using `ApcCache` to run on PHP 7. For no real gain as APCu is basically APC without opcode caching (same API, even same code base...).

dunglas:
@Tobion we can also keep the same PHP signature but remove calls to `apc_*` functions at same time the support for PHP < 5.5 is removed.

layanto:
Agree that if support for PHP < 5.5 will be dropped then replace all apc_* functions with the apcu_* equivalent. Agree with @dunglas re keeping it as ApcCache as APCu is the APC userland cache so not wrong calling it ApcCache. This way no impact on existing code and no need to enable APC backward compatibility feature of APCu.

sroze:
IMO, I would create another cache provider, properly named APCu. That will have the advantage of not disturbing people because APC do not exist anymore and this difference between APC and APCu is a source of confusion a lot of time. That way, the debate about the supported version is also closed :)

dunglas:
What do you think @Ocramius? Do you prefer a new class or keeping this one?

Tobion:
Since support of PHP < 5.5 is going to be dropped in next version as well, we can just use `ApcCache` and change the implementation to use apcu completely.

Ocramius:
@dunglas new class indeed, and deprecate the previous one.

Tobion:
@Ocramius wouldn't it be enough to just change the implementation of ApcCache since PHP < 5.5 support is dropped anyway. So people using PHP < 5.5 cannot use the install the new version anyway and thus there is no bc break. And people that already use PHP >= 5.5 already need APCu, thus there should not be a change for them either.
This way, people do not need to update code and we do not need to add a new config for apcu in doctrine-cache-bundle.

Ocramius:
@Tobion my concern there is that some nuts may still install APC manually, no?

dunglas:
@Ocramius AFAIK the original APC doesn't support any version of PHP >= 5.5.

Ocramius:
I am aware of that, but we'd have to still do feature discovery for those folks that still compile stuff.

Therefore I'd rather have the user opt-in specifically for APCu, rather than some dynamic decisions done by us (especially now that apcu decided to have its own API).

Otherwise, we may conflict with "ext-apc" specifically (which currently breaks the install for APCu as well, no?)

stof:
@Ocramius these guys would have to port the C code to be compatible with newer versions of PHP, so they are really motivated :smile:

Ocramius:
I am aware of that, that's why I used "nuts" ;-)

dunglas:
@Ocramius in the current state of this PR, it can handle both APIs (the APC/BC layer one and the APCu one) using the constructor flag.

Ocramius:
Yes, and that's why I'd prefer separate classes in first place (or dropping `apc` altogether, which depends on #109)

guilhermeblanco:
I'd implement ApcuCache only for PHP7+.
The API did change for the 5.X version of APCu, so we should focus on the non-bc compliant version only, not on older one which has BC compat layer.

sroze:
Thumb up for the new APCu class, as it'll remove the confusion at the same time.

@guilhermeblanco how would you ensure it's only compatible for PHP7+ ? If it's BC with PHP 5.5 is not a big problem IMO :stuck_out_tongue:

guilhermeblanco:
@sroze You didn't get what I said. APCu version 4.X not only supports `apcu_*` functions, but also `apc_*`.
However, on APCu 5.X, there's no BC layer anymore. They created a separate package to handle that (https://github.com/krakjoe/apcu-bc). Still, to properly fix concurrency (read as cache stampede) and not break everything else, they created a new function `apcu_entry` which is only available >= 5.1.0.

My suggestion is that we create an `ApcuCache` class that only supports PHP7+, with `apcu_entry`, and not 5.5 version, because the BC layer would work fine in `ApcCache`. Makes sense now?

sroze:
@guilhermeblanco right, got it. But in order to not break the current `Cache` interface, we can't really use `apcu_entry` as it is, so how would you use it (by keeping the concurrency safe) ?

stof:
@Ocramius given that https://github.com/doctrine/cache/pull/115 was merged, this should probably be closed

PR #114, opened 2015-12-09
Hotfix: mongo collection mocking fixes / 5.3 test failures

Ocramius:


stof:
the APCu installation should be changed, to force it to install 4.x on PHP 5.5 and 5.6 (pecl will try to get the latest even though it requires PHP 7, as it does not have proper dependency resolution)

Ocramius:
@stof yep, figured it out and already pushed

stof:
@Ocramius what about installing APCu 5 for PHP 7 as it is supported now ?

Ocramius:
@stof that already works, but `pecl` is broken on travis-ci atm

PR #115, opened 2015-12-09
Add APCu cache provider

dunglas:
Add native APCu support (replace #112).

@guilhermeblanco, I don't think that it is possible to add `apcu_entry` support without adding a new interface to Doctrine Cache.

### DIFF
--- a/tests/Doctrine/Tests/Common/Cache/ApcuCacheTest.php
+++ b/tests/Doctrine/Tests/Common/Cache/ApcuCacheTest.php
@@ -1,0 +1,21 @@
+<?php
+
+namespace Doctrine\Tests\Common\Cache;
+
+use Doctrine\Common\Cache\ApcuCache;
+
+/**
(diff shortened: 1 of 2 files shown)


tests/Doctrine/Tests/Common/Cache/ApcuCacheTest.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+namespace Doctrine\Tests\Common\Cache;
+
+use Doctrine\Common\Cache\ApcuCache;
+
+/**
+ * @requires extension apcu
```
Ocramius: Should it also skip on PHP < 7.0?

dunglas: It is not necessary as it works. If we add support for apcu_entry later we can skip only this specific feature.

Ocramius:
:+1:

PR #117, opened 2015-12-10
Added MultiPutCache interface and implementations for drivers that su…

danut007ro:
…pports multiput

Serialization and tests

### DIFF
--- a/lib/Doctrine/Common/Cache/ApcCache.php
+++ b/lib/Doctrine/Common/Cache/ApcCache.php
@@ -89,11 +89,7 @@ class ApcCache extends CacheProvider
     {
         $result = apc_store($keysAndValues, null, $lifetime);
 
-        if ($result === false || count($result)) {
-            return false;
-        }
-
-        return true;
+        return empty($result);
     }
 
     /**
--- a/lib/Doctrine/Common/Cache/ApcCache.php
+++ b/lib/Doctrine/Common/Cache/ApcCache.php
@@ -85,6 +85,20 @@ class ApcCache extends CacheProvider
     /**
      * {@inheritdoc}
      */
+    protected function doSaveMultiple(array $keysAndValues, $lifetime = 0)
+    {
+        $result = apc_store($keysAndValues, null, $lifetime);
+
+        if ($result === false || count($result)) {
+            return false;
+        }
+
+        return true;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
     protected function doGetStats()
     {
         $info = apc_cache_info('', true);
--- a/lib/Doctrine/Common/Cache/ApcuCache.php
+++ b/lib/Doctrine/Common/Cache/ApcuCache.php
@@ -80,6 +80,20 @@ class ApcuCache extends CacheProvider
     /**
      * {@inheritdoc}
      */
+    protected function doSaveMultiple(array $keysAndValues, $lifetime = 0)
+    {
+        $result = apcu_store($keysAndValues, null, $lifetime);
+
+        if ($result === false || count($result)) {
+            return false;
+        }
+
+        return true;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
     protected function doGetStats()
     {
         $info = apcu_cache_info(true);
--- a/lib/Doctrine/Common/Cache/ApcuCache.php
+++ b/lib/Doctrine/Common/Cache/ApcuCache.php
@@ -84,11 +84,7 @@ class ApcuCache extends CacheProvider
     {
         $result = apcu_store($keysAndValues, null, $lifetime);
 
-        if ($result === false || count($result)) {
-            return false;
-        }
-
-        return true;
+        return empty($result);
     }
 
     /**
--- a/lib/Doctrine/Common/Cache/CacheProvider.php
+++ b/lib/Doctrine/Common/Cache/CacheProvider.php
@@ -252,7 +252,7 @@ abstract class CacheProvider implements Cache, FlushableCache, ClearableCache, M
     abstract protected function doContains($id);
 
     /**
-     * Default implementation of doSaveMultiple. Each driver that supports multi-put should overwrite it.
+     * Default implementation of doSaveMultiple. Each driver that supports multi-put should override it.
      *
      * @param array $keysAndValues  Array of keys and values to save in cache
      * @param int   $lifetime       The lifetime. If != 0, sets a specific lifetime for these
--- a/lib/Doctrine/Common/Cache/CacheProvider.php
+++ b/lib/Doctrine/Common/Cache/CacheProvider.php
@@ -29,7 +29,7 @@ namespace Doctrine\Common\Cache;
  * @author Roman Borschel <roman@code-factory.org>
  * @author Fabio B. Silva <fabio.bat.silva@gmail.com>
  */
-abstract class CacheProvider implements Cache, FlushableCache, ClearableCache, MultiGetCache
+abstract class CacheProvider implements Cache, FlushableCache, ClearableCache, MultiGetCache, MultiPutCache
 {
     const DOCTRINE_NAMESPACE_CACHEKEY = 'DoctrineNamespaceCacheKey[%s]';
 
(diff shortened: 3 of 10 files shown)


Ocramius:
@danut007ro seems simple/solid to me.
@guilhermeblanco is there any location in the ORM where we can use this?

kimhemsoe:
@Ocramius Maybe the redis providers could use multi or pipeline?

stof:
@Ocramius the SLC should be review to see whether it can benefit from it (this is the place benefiting from MultiGet already)

lib/Doctrine/Common/Cache/ApcCache.php
```diff
@@ -85,6 +85,20 @@ protected function doFetchMultiple(array $keys)
     /**
      * {@inheritdoc}
      */
+    protected function doSaveMultiple(array $keysAndValues, $lifetime = 0)
+    {
+        $result = apc_store($keysAndValues, null, $lifetime);
+
+        if ($result === false || count($result)) {
```
Ocramius: Checking against `false` seems wrong, as you will always get an `array` result, no?
Also, you should probably just use `empty()`, to avoid counting the elements when not needed.

lib/Doctrine/Common/Cache/ApcuCache.php
```diff
@@ -80,6 +80,20 @@ protected function doFetchMultiple(array $keys)
     /**
      * {@inheritdoc}
      */
+    protected function doSaveMultiple(array $keysAndValues, $lifetime = 0)
+    {
+        $result = apcu_store($keysAndValues, null, $lifetime);
+
+        if ($result === false || count($result)) {
```
Ocramius: Same as above

lib/Doctrine/Common/Cache/CacheProvider.php
```diff
@@ -239,6 +252,28 @@ protected function doFetchMultiple(array $keys)
     abstract protected function doContains($id);
 
     /**
+     * Default implementation of doSaveMultiple. Each driver that supports multi-put should overwrite it.
```
Ocramius: s/overwrite/override

lib/Doctrine/Common/Cache/MemcachedCache.php
```diff
@@ -80,6 +80,18 @@ protected function doFetchMultiple(array $keys)
     /**
      * {@inheritdoc}
      */
+    protected function doSaveMultiple(array $keysAndValues, $lifetime = 0)
+    {
+        if ($lifetime > 30 * 24 * 3600) {
```
Ocramius: I'm confused by this statement, but I see that it is replicated in `doSave` :-\

lib/Doctrine/Common/Cache/MultiPutCache.php
```diff
@@ -0,0 +1,41 @@
...
+ * <http://www.doctrine-project.org>.
+ */
+
+namespace Doctrine\Common\Cache;
+
+/**
+ * Interface for cache drivers that allows to put many items at once.
+ *
+ * @link   www.doctrine-project.org
+ * @since  1.4
```
Ocramius: `1.6`

lib/Doctrine/Common/Cache/PredisCache.php
```diff
@@ -52,6 +52,34 @@ protected function doFetchMultiple(array $keys)
...
+    {
+        $success = true;
+
+        if (!$lifetime) {
+            // No lifetime, use MSET
+            $response = $this->client->mset(array_map(function ($value) {
+                return serialize($value);
+            }, $keysAndValues));
+
+            $success = ((string) $response === 'OK');
```
Ocramius: Return early

Ocramius: If the response is `OK`, then there is no need for a cast (strict comparison is already dealing with that)

lib/Doctrine/Common/Cache/PredisCache.php
```diff
@@ -52,6 +52,34 @@ protected function doFetchMultiple(array $keys)
     /**
      * {@inheritdoc}
      */
+    protected function doSaveMultiple(array $keysAndValues, $lifetime = 0)
+    {
+        $success = true;
+
+        if (!$lifetime) {
```
Ocramius: If you remove the `!` you can simply swap the two code blocks

lib/Doctrine/Common/Cache/CacheProvider.php
```diff
@@ -239,6 +252,28 @@ protected function doFetchMultiple(array $keys)
...
+     *
+     * @return bool TRUE if the operation was successful, FALSE if it wasn't.
+     */
+    protected function doSaveMultiple(array $keysAndValues, $lifetime = 0)
+    {
+        $success = true;
+
+        foreach ($keysAndValues as $key => $value) {
+            if (!$this->doSave($key, $value, $lifetime)) {
+                $success = false;
```
Ocramius: This logic assumes that the cache is not going to bail out early when saving an item fails: this requires a test case with a failure.

I suggest coding this sort of test case against a mocked version of the `CacheProvider` that implements `doSave` that triggers a failure

lib/Doctrine/Common/Cache/RedisCache.php
```diff
@@ -88,6 +88,28 @@ protected function doFetchMultiple(array $keys)
     /**
      * {@inheritdoc}
      */
+    protected function doSaveMultiple(array $keysAndValues, $lifetime = 0)
+    {
+        $success = true;
+
+        if (!$lifetime) {
+            // No lifetime, use MSET
+            $success = $this->redis->mset($keysAndValues);
```
Ocramius: Return early.

Ocramius: Actually, please make sure `mset` returns a boolean as well

lib/Doctrine/Common/Cache/RedisCache.php
```diff
@@ -88,6 +88,28 @@ protected function doFetchMultiple(array $keys)
     /**
      * {@inheritdoc}
      */
+    protected function doSaveMultiple(array $keysAndValues, $lifetime = 0)
+    {
+        $success = true;
+
+        if (!$lifetime) {
```
Ocramius: Don't swap this conditional

lib/Doctrine/Common/Cache/WinCacheCache.php
```diff
@@ -83,6 +83,20 @@ protected function doFetchMultiple(array $keys)
     /**
      * {@inheritdoc}
      */
+    protected function doSaveMultiple(array $keysAndValues, $lifetime = 0)
+    {
+        $result = wincache_ucache_set($keysAndValues, null, $lifetime);
+
+        if ($result === false || count($result)) {
```
Ocramius: Same as above: rely on `empty()` and don't check for `false` if an array is returned

Tobion:
I don't think it's a good idea to introduce a new interface for this. IMO MultiGetCache should be merged into Cache in 2.0 anyway. CacheProvider already emulates fetchMultiple anyway which is always possible. And the same would be true for saveMultiple.
So IMO this features should be implemented without an interface. Otherwise we would just create an interface that we don't want in the long run at all.

Ocramius:
@Tobion a few things:
 * changing an existing interface needs a strong reason to do so
 * save and load can already be on separate interfaces: they are not necessarily symmetrical operations anyway
 * this patch is in line with already quite successful interface segregation of this component: I'd rather keep it like that, then aggregate all interfaces into something like `CacheProvider` for 3.x

From my point of view, the patch is ok, but indeed we may re-design some of the interfaces for 3.x

Tobion:
Well if ISP is really wanted for such a small domain, then it should be done fully. Currently it's just random and more based on historic reasons than on ISP. For example, why the `getStats` in Cache when it is probably the least common used method. And if multi save and multi get are different interfaces than, single save and single get should be too. Otherwise it's arbitrary.

Ocramius:
@Tobion fully agree. I think `Zend\Cache` does it quite right on that front (each adapter has a series of capabilities)

danut007ro:
@Ocramius I updated the code according to your review (thank you for it). Have a look and tell me what you think. Thanks!

Ocramius:
Looks good to me.

Waiting for @guilhermeblanco to do a last sweep.

andrerom:
Nice, this will become useful.

Ocramius:
got a :+1: from @guilhermeblanco! Merging \o/ 

Thanks @danut007ro

PR #122, opened 2015-12-17
Hotfix - #113 testing/correcting hashing for windows file path length limitations

Ocramius:
Provides tests for #113

PR #125, opened 2015-12-18
Fix FileCache on Windows

### LINKED ISSUE #124: Filenames are too long for windows
Caching was not working for me properly on my computer so I tried to dig up why. Turns out that this [rename](https://github.com/doctrine/cache/blob/fb8eaa0b612b543d10f09df8eb16b47dd976e97a/lib/Doctrine/Common/Cache/FileCache.php#L254) call often fails with `Warning rename(...) Invalid argument`. Of course the error was suppressed by the `@` operator so it took a while to find. The warning happens when the path is too long (probably 256 characters is a limit).

I tried to upgrade to dev-master, it didn't help.

enumag:
Trying to fix #124.


zeroedin-bill:
See https://github.com/doctrine/cache/pull/122

PR #127, opened 2015-12-19
Path length of 259 is also not possible due to php bug

Tobion:
https://github.com/doctrine/cache/pull/107#issuecomment-165910117
https://bugs.php.net/bug.php?id=70943


Tobion:
@zeroedin-bill you were a little to early as it's still not working. You can create a new tag right away. ;)

zeroedin-bill:
FML WINDOWS

zeroedin-bill:
Thanks, Tobion - Should be solidly fixed now. Please take a look if you have a moment and let me know if you see any further issues.

PR #130, opened 2015-12-21
Added support for stats and ttl on ArrayCache

victorjonsson:
I had some long-time running scripts that was heavily using the ArrayCache. After a while I got some unexpected behaviour, and at first I couldn't figure out why. Imagine my surprise when I discovered that the ttl-feature was missing, it explained a lot.

So here you have my take on how it could be implemented. I also added support for hit ratio. Let me know what you think!

### DIFF
--- a/lib/Doctrine/Common/Cache/ArrayCache.php
+++ b/lib/Doctrine/Common/Cache/ArrayCache.php
@@ -48,7 +48,7 @@ class ArrayCache extends CacheProvider
     public function __construct()
     {
         $this->stats = array(
-            Cache::STATS_HITS   => 0,
+            Cache::STATS_HITS => 0,
             Cache::STATS_MISSES => 0,
             Cache::STATS_UPTIME => time(),
             Cache::STATS_MEMORY_USAGE => null,
--- a/lib/Doctrine/Common/Cache/ArrayCache.php
+++ b/lib/Doctrine/Common/Cache/ArrayCache.php
@@ -38,11 +38,37 @@ class ArrayCache extends CacheProvider
     private $data = array();
 
     /**
+     * @var array $stats
+     */
+    private $stats = array();
+
+
+    /**
+     * ArrayCache constructor.
+     */
+    public function __construct()
+    {
+        $this->stats = array(
+            Cache::STATS_HITS   => 0,
+            Cache::STATS_MISSES => 0,
+            Cache::STATS_UPTIME => time(),
+            Cache::STATS_MEMORY_USAGE => null,
+            Cache::STATS_MEMORY_AVAILABLE => null,
+        );
+    }
+
+    /**
      * {@inheritdoc}
      */
     protected function doFetch($id)
     {
-        return $this->doContains($id) ? $this->data[$id] : false;
+        if ($this->doContains($id)) {
+            $this->stats[Cache::STATS_HITS]++;
+            return $this->data[$id]['content'];
+        } else {
+            $this->stats[Cache::STATS_MISSES]++;
+            return false;
+        }
     }
 
     /**
@@ -51,7 +77,14 @@ class ArrayCache extends CacheProvider
     protected function doContains($id)
     {
         // isset() is required for performance optimizations, to avoid unnecessary function calls to array_key_exists.
-        return isset($this->data[$id]) || array_key_exists($id, $this->data);
+        if (isset($this->data[$id])) {
+            if ($this->data[$id]['ttl'] !== false && $this->data[$id]['ttl'] < time()) {
+                $this->doDelete($id);
+            } else {
+                return true;
+            }
+        }
+        return false;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


Enleur:
+1, have the same problem, so I had to use external storages and rework code to avoid frequent calls.

lib/Doctrine/Common/Cache/ArrayCache.php
```diff
@@ -59,8 +92,8 @@ protected function doContains($id)
      */
     protected function doSave($id, $data, $lifeTime = 0)
     {
-        $this->data[$id] = $data;
-
+        $ttl = $lifeTime > 0 ? time()+$lifeTime : false;
+        $this->data[$id] = array('content'=>$data, 'ttl'=>$ttl);
```
Tobion: this should use numeric indexes instead of `content` and `ttl` to save memory. or use different arrays for data and ttl.

lib/Doctrine/Common/Cache/ArrayCache.php
```diff
@@ -38,11 +38,37 @@ class ArrayCache extends CacheProvider
     private $data = array();
 
     /**
+     * @var array $stats
+     */
+    private $stats = array();
+
+
```
Tobion: extra line

lib/Doctrine/Common/Cache/ArrayCache.php
```diff
@@ -38,11 +38,37 @@ class ArrayCache extends CacheProvider
...
+    /**
+     * ArrayCache constructor.
+     */
+    public function __construct()
+    {
+        $this->stats = array(
+            Cache::STATS_HITS   => 0,
```
Tobion: no alignment

victorjonsson:
@Tobion I've changed to numeric indexes and removed the extra line. I'm unsure about what you mean by "no alignment", let me know what is incorrect and I'll fix it.

Enleur:
@victorjonsson I'm not quite sure about stats, it can contain large numbers in long running processes or daemons

lib/Doctrine/Common/Cache/ArrayCache.php
```diff
@@ -38,11 +38,36 @@ class ArrayCache extends CacheProvider
...
+    /**
      * {@inheritdoc}
      */
     protected function doFetch($id)
     {
-        return $this->doContains($id) ? $this->data[$id] : false;
+        if ($this->doContains($id)) {
+            $this->stats[Cache::STATS_HITS]++;
+            return $this->data[$id][0];
+        } else {
```
Ocramius: No need for `else`

lib/Doctrine/Common/Cache/ArrayCache.php
```diff
@@ -51,7 +75,15 @@ protected function doFetch($id)
     protected function doContains($id)
     {
         // isset() is required for performance optimizations, to avoid unnecessary function calls to array_key_exists.
-        return isset($this->data[$id]) || array_key_exists($id, $this->data);
+        if (isset($this->data[$id])) {
+            $ttl = $this->data[$id][1];
+            if ($ttl !== false && $ttl < time()) {
+                $this->doDelete($id);
+            } else {
+                return true;
```
Ocramius: Return early, avoid the `else`

lib/Doctrine/Common/Cache/ArrayCache.php
```diff
@@ -38,11 +38,35 @@ class ArrayCache extends CacheProvider
...
+    }
+
+    /**
      * {@inheritdoc}
      */
     protected function doFetch($id)
     {
-        return $this->doContains($id) ? $this->data[$id] : false;
+        if ($this->doContains($id)) {
+            $this->stats[Cache::STATS_HITS]++;
```
Ocramius: Use a separate private property for hit stats, instead of storing it in a multidimensional array

lib/Doctrine/Common/Cache/ArrayCache.php
```diff
@@ -38,11 +38,35 @@ class ArrayCache extends CacheProvider
...
+     */
+    private $stats = array();
+
+    /**
+     * {@inheritdoc}
+     */
+    public function __construct()
+    {
+        $this->stats = array(
+            Cache::STATS_HITS => 0,
```
Ocramius: You should count the hits

lib/Doctrine/Common/Cache/ArrayCache.php
```diff
@@ -38,11 +38,35 @@ class ArrayCache extends CacheProvider
...
+            Cache::STATS_MISSES => 0,
```
Ocramius: You should count the misses

lib/Doctrine/Common/Cache/ArrayCache.php
```diff
@@ -38,11 +38,35 @@ class ArrayCache extends CacheProvider
...
      * {@inheritdoc}
      */
     protected function doFetch($id)
     {
-        return $this->doContains($id) ? $this->data[$id] : false;
+        if ($this->doContains($id)) {
+            $this->stats[Cache::STATS_HITS]++;
+            return $this->data[$id][0];
+        }
+        $this->stats[Cache::STATS_MISSES]++;
```
Ocramius: Use a separate private property for miss stats, instead of storing it in a multidimensional array

Ocramius:
@victorjonsson applied changes manually and merged, thanks!

PR #137, opened 2016-01-15
Fix ApcCache on PHP 7 where only APCu is available

nicolas-grekas:
So that libs/users can upgrade smoothly to PHP 7 without breaking their app

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,8 @@
         {"name": "Johannes Schmitt", "email": "schmittjoh@gmail.com"}
     ],
     "require": {
-        "php": ">=5.3.2"
+        "php": ">=5.3.2",
+        "symfony/polyfill-apcu": "~1.0,>=1.0.2"
     },
     "require-dev": {
         "phpunit/phpunit":         ">=3.7",
--- a/lib/Doctrine/Common/Cache/ApcCache.php
+++ b/lib/Doctrine/Common/Cache/ApcCache.php
@@ -37,7 +37,7 @@ class ApcCache extends CacheProvider
      */
     protected function doFetch($id)
     {
-        return apc_fetch($id);
+        return apcu_fetch($id);
     }
 
     /**
@@ -45,7 +45,7 @@ class ApcCache extends CacheProvider
      */
     protected function doContains($id)
     {
-        return apc_exists($id);
+        return apcu_exists($id);
     }
 
     /**
@@ -53,7 +53,7 @@ class ApcCache extends CacheProvider
      */
     protected function doSave($id, $data, $lifeTime = 0)
     {
-        return (bool) apc_store($id, $data, (int) $lifeTime);
+        return (bool) apcu_store($id, $data, (int) $lifeTime);
     }
 
     /**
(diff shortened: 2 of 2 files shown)


Tobion:
This conflicts with the solution taken in #115

nicolas-grekas:
How does it conflict? To me, it complements...

nicolas-grekas:
The issue with #115 is that it's in a branch where only ~5.5|~7.0 are allowed, thus making transitioning harder.

Tobion:
They decided to not make the transition implicit but explicit. Otherwise they could have chosen the original implementation to use apcu or apc depending on what is available.
Also it now would introduce two classes that both use `apcu` functions which doesnt make sense.

nicolas-grekas:
@Tobion you're talking about implementation details, I'm talking about public interfaces, which what matters when thinking about transitioning.
I'm not suggesting to not deprecate ApcCache. In fact, ApcCache is deprecated in 1.6 in favor of ApcuCache, which is really the way to go and makes this move explicit rather than implicit.
But still, ApcCache does not work on PHP 7 and this needs to be fixed.
This PR does the job.

stof:
@nicolas-grekas it can work on PHP 7 depending on how you install apcu (they kept the apc BC layer, it is just disabled by default)

nicolas-grekas:
@stof right. Still, I think this PR makes transitioning a bit nicer/easier. Any downside?

Tobion:
I'm not talking about implementation details. This approach (that ApcCache works both with apcu and apc) was basically already proposed but declined by @Ocramius.

nicolas-grekas:
@Tobion OK, then I'd like this to be reconsidered in the light of the new symfony/polyfill-apcu. It would really help with the transition period.

nicolas-grekas:
ping @Ocramius

guilhermeblanco:
-1 on this approach. If the transition of users depends on only switching ApcCache to ApcuCache, this would take < 5s.

If users are on PHP 5.5, you just added a new layer of indirection (another HashTable lookup) for absolutely no reason.

GrahamCampbell:
:-1: for making any changes to 1.4.

nicolas-grekas:
In Symfony 2.7, we use ApcCache, and we can't move to ApcuCache because our minimum requirement is PHP 5.3.9, which is incompatible with your composer.json v1.6, who requires PHP 5.5
Using polyfill-apcu won't help either because ApcCache implementation is based on `apc_*` calls.
This means Symfony 2.7 currently can't use apcu without the apc layer being enabled.
This is what I call a broken transitioning.
A fixed one would allow to move smoothly from apc to apcu+apc to apcu only, which this PR provides.

Ocramius:
A new specific adapter [was introduced to support APCu](https://github.com/doctrine/cache/blob/e0ef9e9833ca024d251e77e8d82a00ca92b0aa71/lib/Doctrine/Common/Cache/ApcuCache.php) (#115)

Users relying on APC and upgrading to PHP 7 should simply move to 1.6 and rely on the new cache adapter after switching to APCu.

Also, unless security issues come up, 1.4.x is closed for patches, sorry.

PR #138, opened 2016-01-16
APC skip & fix fetchMultiple on failure

Tobion:
1. skip apc tests if apc.enable_cli is not enabled instead of failing (unfortunately this setting cannot be enabled at runtime)
2. fix fetchMultiple implementations that can return false on failure instead of an array. I realized this when looking at the test results in #136

> 1) Doctrine\Tests\Common\Cache\ApcCacheTest::testFetchMultiple
array_key_exists() expects parameter 2 to be array, boolean given
C:\Users\fr-mktg-od\Desktop\Sites\cache\lib\Doctrine\Common\Cache\CacheProvider.php:98

### DIFF
--- a/lib/Doctrine/Common/Cache/ApcCache.php
+++ b/lib/Doctrine/Common/Cache/ApcCache.php
@@ -79,7 +79,7 @@ class ApcCache extends CacheProvider
      */
     protected function doFetchMultiple(array $keys)
     {
-        return apc_fetch($keys);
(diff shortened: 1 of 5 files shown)


Tobion:
@Ocramius ping

Ocramius:
@Tobion thanks! Merged and backported into `1.6.x` via doctrine/cache@71bdc0ddcd5837fcb45b3c1c0e089dce9220790a

Issue #139, opened 2016-01-21
AutoExpiration index of MongoCollections not configured automatically


holtkamp:
After switching to Doctrine\Cache using the MongoDBCache I noticed the MongoDB Collections were growing / expired entries were not removed automatically.

After having a look at the source, it seems the use of [MongoDB's automatic purging functionality is not 'enforced'](https://github.com/doctrine/cache/blob/e0ef9e9833ca024d251e77e8d82a00ca92b0aa71/lib/Doctrine/Common/Cache/MongoDBCache.php#L44-L48). Would it make sense to do this? Something like this which can be invoked in the constructor.

``` php
    private function createIndex()
    {
        $this->collection->createIndex(
            array(self::EXPIRATION_FIELD => 1),
            array(
                'background' => true, //Builds the index in the background so that building an index does not block other database activities.
                'expireAfterSeconds' => 0 // Have entries expire directly (0 seconds) after reaching expiration time
            )
        );
    }
```

Note that prior to MongoDB the function was called [ensureIndex()](http://php.net/manual/en/mongocollection.ensureindex.php), which is now an alias of [createIndex()](http://php.net/manual/en/mongocollection.createindex.php). Ensuring an index exists sounds more appropriate to me, but hey, it is deprecated :smile: 

Also see our previous cache driver: https://github.com/AntonStoeckl/Zend_Cache_Backend_Mongo/blob/aa1ac7cf415bb96fd114a9717e5ab88822462ed5/library/Zend/Cache/Backend/Mongo.php#L548-L567

Ocramius:
Yes, that makes sense /cc @alcaeus

alcaeus:
Yes, using that functionality makes sense ;)

holtkamp:
@Ocramius , you labeled it as a bug, not sure whether that is the case, more new functionality?

I can come up with a PR, might take some weeks before I can find time though...

Ocramius:
> @Ocramius , you labeled it as a bug, not sure whether that is the case, more new functionality?

Because of 

 > I noticed the MongoDB Collections were growing / expired entries were not removed automatically.

PR #140, opened 2016-01-21
Ensure an index is set on the expiration field to enable automatic pu…

holtkamp:
As suggested in https://github.com/doctrine/cache/issues/139

### DIFF
--- a/lib/Doctrine/Common/Cache/MongoDBCache.php
+++ b/lib/Doctrine/Common/Cache/MongoDBCache.php
@@ -75,6 +75,25 @@ class MongoDBCache extends CacheProvider
     public function __construct(MongoCollection $collection)
     {
         $this->collection = $collection;
+        $this->createIndex();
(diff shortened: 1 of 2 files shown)


holtkamp:
Small ping on this, anything missing on having this merged?

Tobion:
This will do an additional MongoDB request every time. Does not seem optimal. IMO this should be done only explicitly as it is a one time setup basically.

Ocramius:
Pretty much agree with @Tobion here, but unsure on how to proceed on this.

holtkamp:
agree with @Tobion as well, but not sure whether "only do this when required" is possible. Might be better as a configurable feature... 

Maybe a small test to see how this impacts performance makes sense? In my case the benefits outweigh the disadvantages 😉

alcaeus:
Superseded by #221. Thanks @holtkamp for getting this started.

PR #146, opened 2016-03-11
Fix save multiple on memcached

rgomezcasas:
The `setMulti` signature for memcached is http://php.net/manual/en/memcached.setmulti.php:
```php
public bool Memcached::setMulti ( array $items [, int $expiration ] )
```
But there are more... In the extension definition (https://github.com/php-memcached-dev/php-memcached/blob/master/php_memcached.c#L1252) this has a third param.

This fixes passing the `$lifetime` instead of the `udf_flags`.

### DIFF
--- a/lib/Doctrine/Common/Cache/MemcachedCache.php
+++ b/lib/Doctrine/Common/Cache/MemcachedCache.php
@@ -86,7 +86,7 @@ class MemcachedCache extends CacheProvider
             $lifetime = time() + $lifetime;
         }
 
-        return $this->memcached->setMulti($keysAndValues, null, $lifetime);
+        return $this->memcached->setMulti($keysAndValues, $lifetime);
     }
 
     /**


jorgeavilacardenosa:
:+1:

Ocramius:
Did the interface change over time, or is this an actual bug affecting all versions of memcached?

rgomezcasas:
Hey, I think this affects all versions of memcached (https://github.com/php-memcached-dev/php-memcached/blob/v1.0.2/php_memcached.c#L912)

I see also this method in doctrine/cache was added 3 month ago: https://github.com/doctrine/cache/commit/62fe59a72a905b9855ce1c92eab278073a87e868#diff-3225c55f6c568cf8f820c5489845ad9bR89

Ocramius:
@rgomezcasas thanks! I backported this into 1.6.x via doctrine/cache@fe3020b50c70ced4562edc04cdf8d0857008502b

PR #148, opened 2016-03-19
Redis::doSaveMultiple: Changed the condition of using the lifetime argument

masterklavi:
It is necessary to use the new condition because the old one was TRUE when a negative value passed.


Ocramius:
While this is not technically a breaking change, it may lead some "WTF" moments during updates, therefore I'm pushing it back to `1.7.0` and closing here, moving to #184

Ocramius:
@masterklavi I ported over your commits, by the way. 👍

PR #149, opened 2016-03-19
Added MemcacheCache::doFetchMultiple() method

masterklavi:
Memcache can to fetch the values by multiple keys and it doesn't need to use doContains, because Memcache doesn't fetch the non-existent values.

### DIFF
--- a/lib/Doctrine/Common/Cache/MemcacheCache.php
+++ b/lib/Doctrine/Common/Cache/MemcacheCache.php
@@ -72,6 +72,14 @@ class MemcacheCache extends CacheProvider
     /**
      * {@inheritdoc}
      */
+    protected function doFetchMultiple(array $keys)
+    {
+        return $this->memcache->get($keys);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
     protected function doContains($id)
     {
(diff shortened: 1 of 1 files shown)


lib/Doctrine/Common/Cache/MemcacheCache.php
```diff
@@ -72,6 +72,14 @@ protected function doFetch($id)
     /**
      * {@inheritdoc}
      */
+    protected function doFetchMultiple(array $keys)
+    {
+        return $this->memcache->get($keys);
```
Ocramius, line 77: What about scenarios where not all of the requested keys can be found?

masterklavi: `doFetchMultiple(array()) === array();`

`doFetchMultiple(array('existing_A', 'existing_B')) === array('existing_A' => ..., 'existing_A' => ...);`

`doFetchMultiple(array('notexisting_A', 'notexisting_B')) === array();`

`doFetchMultiple(array('existing_A', 'notexisting_B')) === array('existing_A' => ...);`

alcaeus:

Issue #147, opened 2016-03-11
FileCache Notice


Stajor:
Hi.
I get a lot of notices in my error log from file vendor/doctrine/cache/lib/Doctrine/Common/Cache/PhpFileCache.php

I saw yours note 
```
note: error suppression is still faster than `file_exists`, `is_file` and `is_readable`
```

But may be you can add file_exists check optional?

Ocramius:
Where do the notices come from? Any details? Can this be reproduced in the test suite?

Stajor:
```
set_error_handler('my_error_handler');
function my_error_handler($errno, $errstr, $errfile, $errlin, $errcontext) {
   // write to error log
}
```
PS.
php 7.0.4

Ocramius:
Right, and can you make that handler fail inside the test suite as well? We can probably avoid error suppression by stacking a new error handler in the execution of the cache fetch logic...

Stajor:
Easy way to reproduce warning
```
        $cache = new \Doctrine\Common\Cache\PhpFileCache('/path_to_cache_folder/');
        set_error_handler(function($errno, $errstr, $errfile, $errlin){
            exit('Oops '.$errstr);
        });
        $cache->fetch('key');
```

Ocramius:
Yep, I'm asking for an actual test case to be added to the test suite ;-)

PR #152, opened 2016-03-27
FileCache Notice

jeanCarloMachado:
Related to https://github.com/doctrine/cache/issues/147

### DIFF
--- a/lib/Doctrine/Common/Cache/PhpFileCache.php
+++ b/lib/Doctrine/Common/Cache/PhpFileCache.php
@@ -108,8 +108,7 @@ class PhpFileCache extends FileCache
     {
         $fileName = $this->getFilename($id);
 
-        // note: error suppression is still faster than `file_exists`, `is_file` and `is_readable`
-        $value = file_exists($fileName) ? include $fileName : null;
+        $value = file_exists($fileName) ? @include $fileName : null;
 
         if (! isset($value['lifetime'])) {
             return false;
--- a/lib/Doctrine/Common/Cache/PhpFileCache.php
+++ b/lib/Doctrine/Common/Cache/PhpFileCache.php
@@ -109,7 +109,7 @@ class PhpFileCache extends FileCache
         $fileName = $this->getFilename($id);
 
         // note: error suppression is still faster than `file_exists`, `is_file` and `is_readable`
-        $value = @include $fileName;
+        $value = file_exists($fileName) ? include $fileName : null;
 
         if (! isset($value['lifetime'])) {
             return false;
--- a/tests/Doctrine/Tests/Common/Cache/PhpFileCacheTest.php
+++ b/tests/Doctrine/Tests/Common/Cache/PhpFileCacheTest.php
@@ -66,6 +66,17 @@ class PhpFileCacheTest extends BaseFileCacheTest
     {
         return new PhpFileCache($this->directory);
     }
+
+    public function testFailureOnIncludeNotSettedCacheDoesntGetCaughtAsError()
+    {
+        $cache = $this->_getCacheDriver();
+        set_error_handler(function($errno, $errstr, $errfile, $errlin){
+            exit('Oops '.$errstr);
+        });
+        $cache->fetch('key');
+
+        restore_error_handler();
+    }
 }
 
 class NotSetStateClass
--- a/tests/Doctrine/Tests/Common/Cache/PhpFileCacheTest.php
+++ b/tests/Doctrine/Tests/Common/Cache/PhpFileCacheTest.php
@@ -70,7 +70,8 @@ class PhpFileCacheTest extends BaseFileCacheTest
     public function testFailureOnIncludeNotSettedCacheDoesntGetCaughtAsError()
     {
         $cache = $this->_getCacheDriver();
-        set_error_handler(function($errno, $errstr, $errfile, $errlin){
+        set_error_handler(function($errno, $errstr, $errfile, $errline){
+            restore_error_handler();
             $this->fail('include failure captured');
         });
         $cache->fetch('key');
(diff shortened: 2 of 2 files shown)


lib/Doctrine/Common/Cache/PhpFileCache.php
```diff
@@ -109,7 +109,7 @@ private function includeFileForId($id)
         $fileName = $this->getFilename($id);
 
         // note: error suppression is still faster than `file_exists`, `is_file` and `is_readable`
-        $value = @include $fileName;
+        $value = file_exists($fileName) ? include $fileName : null;
```
Tobion: the error supression must be kept to avoid race conditions

lib/Doctrine/Common/Cache/PhpFileCache.php
```diff
@@ -109,7 +109,7 @@ private function includeFileForId($id)
         $fileName = $this->getFilename($id);
 
         // note: error suppression is still faster than `file_exists`, `is_file` and `is_readable`
```
Tobion: this note doesn't make sense anymore

tests/Doctrine/Tests/Common/Cache/PhpFileCacheTest.php
```diff
@@ -66,6 +66,17 @@ protected function _getCacheDriver()
     {
         return new PhpFileCache($this->directory);
     }
+
+    public function testFailureOnIncludeNotSettedCacheDoesntGetCaughtAsError()
+    {
+        $cache = $this->_getCacheDriver();
+        set_error_handler(function($errno, $errstr, $errfile, $errlin){
+            exit('Oops '.$errstr);
```
Tobion: this is bad

lib/Doctrine/Common/Cache/PhpFileCache.php
jeanCarloMachado: I'll apply this corrections.

tests/Doctrine/Tests/Common/Cache/PhpFileCacheTest.php
jeanCarloMachado: I agree, I may look for alternatives, you have some?

jeanCarloMachado:
I made the corrections, any other suggestion @Tobion ?

tests/Doctrine/Tests/Common/Cache/PhpFileCacheTest.php
```diff
@@ -66,6 +66,17 @@ protected function _getCacheDriver()
     {
         return new PhpFileCache($this->directory);
     }
+
+    public function testFailureOnIncludeNotSettedCacheDoesntGetCaughtAsError()
+    {
+        $cache = $this->_getCacheDriver();
+        set_error_handler(function($errno, $errstr, $errfile, $errlin){
+            $this->fail('include failure captured');
```
trakos: If that will be triggered restore_error_handler won't be called, causing many other tests to fail as well.

tests/Doctrine/Tests/Common/Cache/PhpFileCacheTest.php
```diff
@@ -66,6 +66,17 @@ protected function _getCacheDriver()
...
+        });
+        $cache->fetch('key');
```
trakos: Maybe test for ``$cache->delete('not_existing_key');`` could also be added - it has the same problem

lib/Doctrine/Common/Cache/PhpFileCache.php
```diff
@@ -108,8 +108,7 @@ private function includeFileForId($id)
     {
         $fileName = $this->getFilename($id);
 
-        // note: error suppression is still faster than `file_exists`, `is_file` and `is_readable`
-        $value = @include $fileName;
+        $value = file_exists($fileName) ? @include $fileName : null;
```
trakos: If error suppression really is faster than file_exists (which would make sense) maybe better way would be:

```php
set_error_handler(function ($errno, $errstr, $errfile, $errline) { /* do nothing */ });
$value = include $fileName;
restore_error_handler();
```
? It will also ignore errors from included file, but that shouldn't make any difference.

tests/Doctrine/Tests/Common/Cache/PhpFileCacheTest.php
jeanCarloMachado: Nice, I'll do that.

lib/Doctrine/Common/Cache/PhpFileCache.php
jeanCarloMachado: It seems that dealing with error_handlers is more expensive than doing a file_exists.
https://3v4l.org/gOXjR
Either way, I'm not sure this kind of micro-optimization is healthy to the code.

Ocramius: Please run benchmarks against something like phpbench/phpbench, if you want
some accurate results

Marco Pivetta

http://twitter.com/Ocramius

http://ocramius.github.com/

tests/Doctrine/Tests/Common/Cache/PhpFileCacheTest.php
jeanCarloMachado: I'll restore the error handler on the capture, it seems to work https://3v4l.org/hn5UC.

tests/Doctrine/Tests/Common/Cache/PhpFileCacheTest.php
```diff
@@ -66,6 +66,18 @@ protected function _getCacheDriver()
     {
         return new PhpFileCache($this->directory);
     }
+
+    public function testFailureOnIncludeNotSettedCacheDoesntGetCaughtAsError()
```
Tobion: You could add this test to the generic CacheTest so it is run against every cache driver. If doctrine decides to support custom error handlers that do not respect the `error_reporting`, then it applies to all drivers and not just the PhpFileCache.

Ocramius:
Moved to #185

PR #154, opened 2016-04-11
Cache anything with PhpFileCache

trakos:
Currently PhpFileCache is a bit useless for doctrine, since it can't cache variables without __set_state implemented. Even MetadataInfo doesn't implement __set_state iirc. The error recommends using FilesystemCache instead, but it's much slower.

Simply adding support for serializing variable before var_export makes this class usable again, and it performs better than FilesystemCache.

I think there is no reason to avoid serialize at all cost, as other cache methods use it anyway.

### DIFF
--- a/lib/Doctrine/Common/Cache/PhpFileCache.php
+++ b/lib/Doctrine/Common/Cache/PhpFileCache.php
@@ -85,12 +85,12 @@ class PhpFileCache extends FileCache
             'data'      => $data
         );
 
-        if (is_object($data) && ! method_exists($data, '__set_state')) {
-            $value  = var_export(serialize($value), true);
-            $code   = sprintf('<?php return unserialize(%s);', $value);
-        } else {
+        if (is_object($data) && method_exists($data, '__set_state')) {
             $value  = var_export($value, true);
             $code   = sprintf('<?php return %s;', $value);
+        } else {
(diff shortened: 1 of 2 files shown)


lib/Doctrine/Common/Cache/PhpFileCache.php
```diff
@@ -78,23 +78,20 @@ protected function doSave($id, $data, $lifeTime = 0)
...
             'data'      => $data
         );
 
-        $value  = var_export($value, true);
-        $code   = sprintf('<?php return %s;', $value);
+        if (is_object($data) && method_exists($data, '__set_state')) {
+            $value  = var_export($value, true);
+            $code   = sprintf('<?php return %s;', $value);
+        } else {
+            $value  = var_export(serialize($value), true);
```
flack: If it were up to me, I'd open a ticket for the array of objects problem and merge this, as it adds support for a new use case without changing anything related to the problem. But like I said, not my PR :-)

trakos: Actually , this code calls `serialize` for an array, it only does simple `var_export` for object implementing `__set_state`. I did this way because I didn't want to traverse an array to check values' types.

Ocramius:
Moved to #190 (rebased/cleaned up a bit). Closing here, thanks @trakos!

PR #156, opened 2016-04-13
Bug fix: treating null lifetime as infinite in PhpFileCache

trakos:
According to docs ( https://github.com/doctrine/doctrine2/blob/13473e8b4ed5dfe6fe1d3a053f8ab8d22752cfe7/docs/en/reference/caching.rst#saving ), null $lifetime should mean infinite lifetime. However, PhpFileCache does isset($value['lifetime']) to determine whether value exists (which returns false for null).

This PR fixes that bug and adds test with null lifetime.

### DIFF
--- a/lib/Doctrine/Common/Cache/CacheProvider.php
+++ b/lib/Doctrine/Common/Cache/CacheProvider.php
@@ -129,7 +129,7 @@ abstract class CacheProvider implements Cache, FlushableCache, ClearableCache, M
      */
     public function save($id, $data, $lifeTime = 0)
     {
-        return $this->doSave($this->getNamespacedId($id), $data, $lifeTime);
+        return $this->doSave($this->getNamespacedId($id), $data, (int) $lifeTime);
     }
 
     /**
@@ -265,7 +265,7 @@ abstract class CacheProvider implements Cache, FlushableCache, ClearableCache, M
         $success = true;
 
         foreach ($keysAndValues as $key => $value) {
-            if (!$this->doSave($key, $value, $lifetime)) {
+            if (!$this->doSave($key, $value, (int) $lifetime)) {
                 $success = false;
             }
         }
--- a/lib/Doctrine/Common/Cache/CouchbaseCache.php
+++ b/lib/Doctrine/Common/Cache/CouchbaseCache.php
@@ -81,7 +81,7 @@ class CouchbaseCache extends CacheProvider
         if ($lifeTime > 30 * 24 * 3600) {
             $lifeTime = time() + $lifeTime;
         }
-        return $this->couchbase->set($id, $data, (int) $lifeTime);
+        return $this->couchbase->set($id, $data, $lifeTime);
     }
 
     /**
--- a/lib/Doctrine/Common/Cache/MemcacheCache.php
+++ b/lib/Doctrine/Common/Cache/MemcacheCache.php
@@ -89,7 +89,7 @@ class MemcacheCache extends CacheProvider
         if ($lifeTime > 30 * 24 * 3600) {
             $lifeTime = time() + $lifeTime;
         }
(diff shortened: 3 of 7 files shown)


Tobion:
The documentation seems wrong. Infinite lifetime is expressed as zero (0) nor null. See https://github.com/doctrine/cache/blob/master/lib/Doctrine/Common/Cache/Cache.php#L74

Tobion:
Also `save($id, $data, $lifeTime = false) ` is also wrong in the doc. Lifetime is not boolean false.

trakos:
Doctrine itself uses null when caching annotations: [AbstractClassMetadataFactory.php#L217-L221](https://github.com/doctrine/common/blob/773c56e51fc26ae07a47146fae31b320dd6cada3/lib/Doctrine/Common/Persistence/Mapping/AbstractClassMetadataFactory.php#L217-L221)

It's probably not the only place. Anyway, all other cache drivers I've checked allows null (usually by casting lifetime to int), for example:
[CouchbaseCache.php#L84](https://github.com/doctrine/cache/blob/master/lib/Doctrine/Common/Cache/CouchbaseCache.php#L84)
[FilesystemCache.php#L89](https://github.com/doctrine/cache/blob/master/lib/Doctrine/Common/Cache/FilesystemCache.php#L89)
[MemcacheCache.php#L92](https://github.com/doctrine/cache/blob/master/lib/Doctrine/Common/Cache/MemcacheCache.php#L92)
[MemcachedCache.php#L109](https://github.com/doctrine/cache/blob/master/lib/Doctrine/Common/Cache/MemcachedCache.php#L109)
[XcacheCache.php#L56](https://github.com/doctrine/cache/blob/master/lib/Doctrine/Common/Cache/XcacheCache.php#L56)

Array cache even converts 0 to false:
[ArrayCache.php#L104](https://github.com/doctrine/cache/blob/master/lib/Doctrine/Common/Cache/ArrayCache.php#L104)
MongoDB uses null for infinite anyway:
[MongoDBCache.php#L127](https://github.com/doctrine/cache/blob/master/lib/Doctrine/Common/Cache/MongoDBCache.php#L127)

I'm pretty sure other will accept null as well. 

Anyway, I can add triggering E_DEPRECATED when null, but I don't think it's worth it.

Tobion:
IMO the int casting should be done in https://github.com/doctrine/cache/blob/master/lib/Doctrine/Common/Cache/CacheProvider.php#L132 and https://github.com/doctrine/cache/blob/master/lib/Doctrine/Common/Cache/CacheProvider.php#L268
Then all implementing cache drivers can be sure to receive an int.

trakos:
Good call, I've changed it.

Tobion:
Please remove the int casting from CouchbaseCache, MemcacheCache, MemcachedCache and XcacheCache then. (Not from FilesystemCache as this is at reading time, not writing).

trakos:
I've pushed the changes, though I don't think it makes much of a difference.

Tobion:
:+1:

Tobion:
@Ocramius ping

Ocramius:
To be checked against #184 discussion. Assigning to others for review.

jindrichpilar-kosik:
This bug/issue causes https://github.com/doctrine/orm/issues/8197. 

Apart from casting, this might be fixed by changing `! isset` to `! is_array || ! array_key_exists` at https://github.com/doctrine/cache/blob/418e2daaa1e7893161f5103bba80b61d1b8be33a/lib/Doctrine/Common/Cache/PhpFileCache.php#L112

stof:
@jindrichpilar-mf you linked to this same PR, not to something in doctrine/orm

jindrichpilar-kosik:
@stof  Thanks for pointing it out, I edited my previous post. I meant https://github.com/doctrine/orm/issues/8197

alcaeus:

PR #163, opened 2016-06-27
Added CacheProvider::fetchAndSave() method

vsmoraes:
I found myself writing this over and over, so I thought it could be a nice addition to the cache provider


stof:
this pattern does not make sense. You are computing the data even when it is already in the cache (to be able to pass it as argument), which means your cache is useless

vsmoraes:
@stof I failed to demonstrate what I really want to achieve. The TL;DR version is that you can pass a closure as `$data`...

I will get back to this PR soon

PR #164, opened 2016-06-28
throw an error if cannot delete

UziTech:
Check if file exists first to remove the error if it does not exists and throw an error for easier debugging if there is a different issue.

this would've saved me hours

### DIFF
--- a/lib/Doctrine/Common/Cache/FileCache.php
+++ b/lib/Doctrine/Common/Cache/FileCache.php
@@ -164,7 +164,7 @@ abstract class FileCache extends CacheProvider
     {
         $filename = $this->getFilename($id);
 
-        return @unlink($filename) || ! file_exists($filename);
+        return ! file_exists($filename) || unlink($filename);
     }
 
     /**
--- a/lib/Doctrine/Common/Cache/FileCache.php
+++ b/lib/Doctrine/Common/Cache/FileCache.php
@@ -164,7 +164,7 @@ abstract class FileCache extends CacheProvider
     {
         $filename = $this->getFilename($id);
 
-        return ! file_exists($filename) || unlink($filename);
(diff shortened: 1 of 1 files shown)


lib/Doctrine/Common/Cache/FileCache.php
```diff
@@ -164,7 +164,7 @@ protected function doDelete($id)
     {
         $filename = $this->getFilename($id);
 
-        return @unlink($filename) || ! file_exists($filename);
```
Ocramius: The idea is to always call `unlink()`, then check if the file exists, should `unlink()` fail.

UziTech: It would be very helpful in debugging to have the error if unlink fails for something other than the file not existing.

Out of curiosity how could `unlink()` fail if it actually deleted the file?

GrahamCampbell: If something else deleted the file first.

GrahamCampbell:
:-1:

lib/Doctrine/Common/Cache/FileCache.php
UziTech: right that's why the check for `file_exists` happens before the `unlink` call

GrahamCampbell: That won't work though, because it good get deleted in-between that cal and the next one.

UziTech: True we could add another check after unlink. That way the functions response will be the same but we will still get an error in the event that something needs to be debugged.

Ocramius:
Same as #152, the check is not needed and is wasteful from an IOPS perspective. Closing as `invalid` therefore.

PR #167, opened 2016-07-15
Fix: exists() returns an integer

localheinz:
This PR

* [x] assumes that `Redis::exists()` returns `0` if the cache doesn't contain data for an `$id`

💁 For reference, see http://redis.io/commands/EXISTS#return-value.


Ocramius:
@localheinz thanks!

PR #168, opened 2016-07-15
Fix: Case-insensitive class references

localheinz:
This PR

* [x] fixes case-insensitive references

### DIFF
--- a/tests/Doctrine/Tests/Common/Cache/WinCacheCacheTest.php
+++ b/tests/Doctrine/Tests/Common/Cache/WinCacheCacheTest.php
@@ -2,15 +2,15 @@
 
 namespace Doctrine\Tests\Common\Cache;
 
-use Doctrine\Common\Cache\WincacheCache;
+use Doctrine\Common\Cache\WinCacheCache;
 
 /**
  * @requires extension wincache
  */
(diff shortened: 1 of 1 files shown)


tests/Doctrine/Tests/Common/Cache/WinCacheCacheTest.php
```diff
@@ -2,15 +2,15 @@
 
 namespace Doctrine\Tests\Common\Cache;
 
-use Doctrine\Common\Cache\WincacheCache;
+use Doctrine\Common\Cache\WinCacheCache;
```
localheinz: @Ocramius 

Reference is case-insensitive.

tests/Doctrine/Tests/Common/Cache/WinCacheCacheTest.php
```diff
@@ -2,15 +2,15 @@
...
 
 /**
  * @requires extension wincache
  */
-class WincacheCacheTest extends CacheTest
+class WinCacheCacheTest extends CacheTest
```
localheinz: @Ocramius 

Class name does not match file name.

Ocramius:
@localheinz thanks!

PR #169, opened 2016-07-15
Fix: Second parameter should probably be an assertion message

localheinz:
This PR

* [x] moves an argument into the right place


Ocramius:
👍

PR #170, opened 2016-07-15
Fix: Class name does not match file name

localheinz:
This PR

* [x] renames a test class so it matches the name of the containing file


Ocramius:
👍

PR #171, opened 2016-07-15
Fix: Use short array syntax

localheinz:
This PR

* [x] uses short array syntax

💁 Ran

```
$ composer global require friendsofphp/php-cs-fixer:2.0.0-alpha
$ php-cs-fixer fix --using-cache=no --rules=short_array_syntax .
```


Ocramius:
👍 🚢

PR #172, opened 2016-07-15
Fix: Remove assignment from condition

localheinz:
This PR

* [x] removes an assignment from a condition

💁 This also makes the happy path more obvious! We're happier if we found something in the cache, aren't we?

### DIFF
--- a/lib/Doctrine/Common/Cache/SQLite3Cache.php
+++ b/lib/Doctrine/Common/Cache/SQLite3Cache.php
@@ -86,11 +86,13 @@ class SQLite3Cache extends CacheProvider
      */
     protected function doFetch($id)
     {
-        if ($item = $this->findById($id)) {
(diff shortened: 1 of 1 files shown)


lib/Doctrine/Common/Cache/SQLite3Cache.php
```diff
@@ -86,11 +86,13 @@ public function __construct(SQLite3 $sqlite, $table)
      */
     protected function doFetch($id)
     {
-        if ($item = $this->findById($id)) {
-            return unserialize($item[self::DATA_FIELD]);
+        $item = $this->findById($id)
```
Ocramius: Whoops, syntax error. Will fixup on master, just a sec

localheinz: @Ocramius 

Ha, sorry! Thank you!

PR #173, opened 2016-07-20
handle invalid namespaceVersion value

vvh-empora:


Ocramius:
Requires test case

Tobion:
When would `getNamespaceVersion` not return an int?

vvh-empora:
i got this error as an buggy PHP7 memcache client...

Tobion:
This workaround doesn't make sense as if that fails, then every other piece you put into the cache could fail too. So I'm :-1: to merge a useless workaround.

Ocramius:
A cache is still to be handled as an external system, @Tobion. Maybe a documentation issue is needed here...

SpacePossum:
>When would getNamespaceVersion not return an int?

the value is fetch from the cache (like @Ocramius stated an external source), any outside process could have written something else to the cache for that key.

IMHO the check should be here btw.:
https://github.com/vvh-empora/cache/blob/c4d131af47c8b101e06d2f737d8326681223678b/lib/Doctrine/Common/Cache/CacheProvider.php#L213

than here:
https://github.com/vvh-empora/cache/blob/c4d131af47c8b101e06d2f737d8326681223678b/lib/Doctrine/Common/Cache/CacheProvider.php#L188

the second `%s` should be `%d`

Tobion:
> the value is fetch from the cache (like @Ocramius stated an external source), any outside process could have written something else to the cache for that key.

As I explaiend above, this also mean ANY another value you set in the cache might have the wrong type. This means for every value you set via the doctrine cache, it would need to also persist the type to ensure when you retrieve it, it's still the expected type.
This is why there are namespaces and if you cannot trust the cache content, then it does not only affect the namespace version but everything which means you have alot more problems than this.

SpacePossum:
That is true, and if you want to deal with you have to do so on in your application, i.e. check what comes back from the cache.
However you cannot deal with the namespace version in your application because you cannot reach it, therefor it should be dealt with here IMO.

alcaeus:

Issue #141, opened 2016-01-27
Misleading cache statistics


BreiteSeite:
The default namespaceVersion (1) is always read, but never written into the cache. See: https://github.com/doctrine/cache/blob/e0ef9e9833ca024d251e77e8d82a00ca92b0aa71/lib/Doctrine/Common/Cache/CacheProvider.php#L212

This increments `get_misses` on memcache on every request.

I would propose to store the default namespaceVersion in the cache too (i think the existing behavior is not clean/logical that we only store versions >1. This makes version 1 a special case and i don't like special cases).

However, there are [tests that explicitly mention that this should not happen](https://github.com/doctrine/cache/blob/master/tests/Doctrine/Tests/Common/Cache/CacheProviderTest.php#L54). Should this assumption be changed and the tests refactored?

If so, i probably can submit a PR.

Thank you for your Feedback.

Tobion:
See arguments in #91

BreiteSeite:
I think these points should get reevaluated because i think the introduced issue was overlooked.

Tobion:
To me performance and space requirements are more important than cache statistics.

Tobion:
Btw, this problem would be solved by #96 anyway as it is a design issue.

BreiteSeite:
Yes.. i'm aware of #96 but i want a simple yet fast "fix" before a huge refactoring hits a release.

However, regarding your points performance and space requirements: i don't believe that the write of the current namespaceVersion into the cache (which is only overhead if the cache is empty) is that much of a performance regression or do you have relevant analytics or benchmarks for this point? Also if you empty your cache, you take a massive performance hit anyways so i'm sure this one additional write into cache doesn't matter.

Also storing an integer value is a joke for space requirements.

Also cache statistics are crucial for performance analysis so i don't see why better statistics is contrary in your eyes to "performance requirements".

Tobion:
This is per namespace. Btw, there is a workaround for you: just call `deleteAll` once. Or set the namespace version yourself using `CacheProvider::DOCTRINE_NAMESPACE_CACHEKEY`.
Also the cache statistics are wrong either way. Either the hits are wrong or the misses.

BreiteSeite:
Your workaround is not valid in my eyes as cache flushing is part of the deployment (no php involved here) and calling deleteAll could erase a freshly populated cache and set the namespace version myself raises also a lot of technical debt (as probably no one after some weeks/months/years will know wtf is going on).

[...]

PR #177, opened 2016-08-11
Save first namespace cache key

nicolas-grekas:
By not saving the "1" namespace, the CacheProvider generates a cache miss at almost each instantiation. Saving this "1" namespace turns this miss into a hit, and thus allows to benefit from the fastest cache when using a ChainCache.


Tobion:
See #141

nicolas-grekas:
Thanks for the link. I'm coding since the blocker is not implementing buy deciding what should be the behavior.
At least here is a new argument that reverses the performance related one.

Issue #150, opened 2016-03-24
Impossible to override Redis serializer


acasademont:
Hi! Even though we have the igbinary extension installed we don't want cache entries to be serialized with igbinary. Unfortunately, even though the phpredis client we use already has the OPT_SERIALIZER set, the RedisCache class ignores that and sets it to use igbinary if it detects the extension is present. Could a method be created to manually set which serializer method we want?

Thanks!

dennis-hh:
+1. We are having issues with igbinary-serialized metadata, and want to use php's serializer instead.

peteward:
Same here. 

We're using a pre-packaged version of php 7.0.6, igbinary and phpredis. The version of phpredis is compiled for igbinary so we cannot simply remove it in our hosting environment.

We see mapping errors whenever class metadata is retrieved from Redis such as:

```
[Doctrine\ORM\Mapping\MappingException]                                                    
  No mapping found for field 'X' on class 'Y'. 
```

or worse:

```
Segmentation Fault
```

The strange thing is that we _only_ see this on console using CLI. Using our web interface works without a problem.

The current workaround for us was to override RedisCache and force using `Redis::SERIALIZER_PHP`.

It feels like the problem here is really with igbinary, but I'm not sure.

Is anyone experiencing the same? I see a similar stories  https://github.com/phpredis/phpredis/issues/762 and https://github.com/snc/SncRedisBundle/issues/259

griesi007:
Same behaviour as explained by @peteward for me - but only if php (PHP 7.0.6, Doctrine 2.5.4) is called from cli. Disabling redis as a doctrine cache backend on cli solves the problem as well as overwriting RedisCache to always use SERIALIZER_PHP.

yellow1912:
`$redis->setOption(Redis::OPT_SERIALIZER, $this->getSerializerValue());`

I think right now the code always set the serializer, this leaves us in situations where we have to use serializer even when we don't want to. I think it should be up to the user to decide this.

ostrolucky:
This would also fix https://github.com/doctrine/common/issues/155

PR #180, opened 2016-10-12
Allow Redis Serializer option to be set outside

yellow1912:
The RedisCache Adapter should not force Serializer Option.
The Serializer Option should be set When the redis client
is created (outside of the scope of this class).

### DIFF
--- a/lib/Doctrine/Common/Cache/RedisCache.php
+++ b/lib/Doctrine/Common/Cache/RedisCache.php
@@ -44,7 +44,11 @@ class RedisCache extends CacheProvider
      */
     public function setRedis(Redis $redis)
     {
-        $redis->setOption(Redis::OPT_SERIALIZER, $this->getSerializerValue());
+        // The redis object is injected from an external service
+        // Serializer option should be set outside
+        // We should not force the serializer option here
+        // This code is left here for reference only
+//        $redis->setOption(Redis::OPT_SERIALIZER, $this->getSerializerValue());
         $this->redis = $redis;
     }
 


lib/Doctrine/Common/Cache/RedisCache.php
```diff
@@ -44,7 +44,11 @@ class RedisCache extends CacheProvider
      */
     public function setRedis(Redis $redis)
     {
-        $redis->setOption(Redis::OPT_SERIALIZER, $this->getSerializerValue());
+        // The redis object is injected from an external service
+        // Serializer option should be set outside
+        // We should not force the serializer option here
+        // This code is left here for reference only
```
stof: code should not be left for reference. It should be removed if it is not used anymore (I cannot tell whether the doctrine/cache maintainers will accept it or no though)

yellow1912: Thanks stof, I will wait for the response from the maintainer then if the change is ok I will remove the reference code as you suggest.

Tobion:
While I agree that it is not be responsibility of the class to set this option (esp. since it cannot be overwritten), changing this is probably a big BC break. Also it makes the method `getSerializerValue` unused and misplaced.

Tobion:
Ref. #150

yellow1912:
Isn't getSerializerValue a protected value only used for this class? Can't we safely drop it?

mvanbaak:
@yellow1912 Can you pick this up? If not, I can create a new PR. (We are affected by this as well, and removing the forcing of the serializer makes the RedisCache in line with MemcachedCache)

mvanbaak:
Created new PR which removes the getSerializerValue method and directly sets the passed Redis object to $this->redis (now in line with MemcachedCache)

yellow1912:
Sorry I was busy. I see you already created new PR, cant wait to see it get merged :)

mvanbaak:
:) I know how that is. Since we are using this fix in production for a week now I thought it was a good idea to share it publicly.

lcobucci:
As discussed in #211 we would break BC with that change.
Closing this to use #211 as recommended solution.

PR #183, opened 2016-10-26
Adding PHP 7.1 to the build matrix

Ocramius:


Ocramius:
Note: fails due to mongo extension failing to load on @travis-ci - possibly to be fixed on that side.

PR #155, opened 2016-04-12
Add MultiDeleteCache interface and default CacheProvider implementation

bburnichon:
I have a case where I need to invalidate several keys at once. This port the addition I made to provide it to all users of this library.

Also add memcached implementation.

PR #186, opened 2016-10-26
Add deleteMultiple support

flack:
Add multi-delete interface and default implementation in the same way as saveMultiple & fetchMultiple are handled. Also implemented in a couple of drivers that support it

### DIFF
--- a/lib/Doctrine/Common/Cache/CacheProvider.php
+++ b/lib/Doctrine/Common/Cache/CacheProvider.php
@@ -28,8 +28,9 @@ namespace Doctrine\Common\Cache;
  * @author Jonathan Wage <jonwage@gmail.com>
  * @author Roman Borschel <roman@code-factory.org>
  * @author Fabio B. Silva <fabio.bat.silva@gmail.com>
+ * @author Andreas Flack <flack@contentcontrol-berlin.de>
  */
-abstract class CacheProvider implements Cache, FlushableCache, ClearableCache, MultiGetCache, MultiPutCache
+abstract class CacheProvider implements Cache, FlushableCache, ClearableCache, MultiGetCache, MultiPutCache, MultiDeleteCache
 {
     const DOCTRINE_NAMESPACE_CACHEKEY = 'DoctrineNamespaceCacheKey[%s]';
 
@@ -86,7 +87,7 @@ abstract class CacheProvider implements Cache, FlushableCache, ClearableCache, M
         if (empty($keys)) {
             return [];
         }
-        
+
         // note: the array_combine() is in place to keep an association between our $keys and the $namespacedKeys
         $namespacedKeys = array_combine($keys, array_map([$this, 'getNamespacedId'], $keys));
         $items          = $this->doFetchMultiple($namespacedKeys);
@@ -114,6 +115,16 @@ abstract class CacheProvider implements Cache, FlushableCache, ClearableCache, M
         }
 
         return $this->doSaveMultiple($namespacedKeysAndValues, $lifetime);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function deleteMultiple(array $keys)
+    {
+        $namespacedKeys = array_map([$this, 'getNamespacedId'], $keys);
+
+        return $this->doDeleteMultiple($namespacedKeys);
     }
 
     /**
@@ -284,6 +295,26 @@ abstract class CacheProvider implements Cache, FlushableCache, ClearableCache, M
      * @return bool TRUE if the entry was successfully stored in the cache, FALSE otherwise.
      */
     abstract protected function doSave($id, $data, $lifeTime = 0);
+
+    /**
+     * Default implementation of doDeleteMultiple. Each driver that supports multi-delete should override it.
+     *
+     * @param array $keys Array of keys to delete from the cache
+     *
+     * @return bool TRUE if the operation was successful, FALSE if it wasn't.
+     */
+    protected function doDeleteMultiple(array $keys)
+    {
+        $success = true;
+
+        foreach ($keys as $key) {
+            if (!$this->doDelete($key)) {
+                $success = false;
+            }
+        }
+
+        return $success;
+    }
 
     /**
      * Deletes a cache entry.
--- a/lib/Doctrine/Common/Cache/MemcachedCache.php
+++ b/lib/Doctrine/Common/Cache/MemcachedCache.php
@@ -92,6 +92,16 @@ class MemcachedCache extends CacheProvider
     /**
      * {@inheritdoc}
      */
+    protected function doDeleteMultiple(array $keys)
+    {
+        $stat = $this->memcached->deleteMulti($keysAndValues, $lifetime);
+        return count(array_unique($stat) > 1)
+            || end($stat) === true;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
     protected function doContains($id)
     {
         return false !== $this->memcached->get($id)
--- a/lib/Doctrine/Common/Cache/MemcachedCache.php
+++ b/lib/Doctrine/Common/Cache/MemcachedCache.php
@@ -94,7 +94,7 @@ class MemcachedCache extends CacheProvider
      */
     protected function doDeleteMultiple(array $keys)
     {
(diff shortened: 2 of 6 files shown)


Ocramius:
@flack can you check versions when the multi-delete was added to apcu, memcached and redis?

Overall, seems good/simple enough.

Ocramius:
Ah, yep, tests missing, sorry... Those are absolutely needed for the feature to be considered.

flack:
I had tested against redis, and now also tested apcu and memcached (good thing, too, since that uncoverd two more problems that are now fixed). Testcases are added as well.

BTW: I'm having some problems with the test suite: For one, it tries to delete my system temp dir (#187), and also, `ApcCache` tests fail (most probably because I don't have `apc` installed, only `apcu`, and it can't tell the difference between the two):

```
There was 1 error:

1) Doctrine\Tests\Common\Cache\ApcCacheTest::testGetStats
Undefined index: nmisses

cache/lib/Doctrine/Common/Cache/ApcCache.php:106
cache/lib/Doctrine/Common/Cache/CacheProvider.php:159
cache/tests/Doctrine/Tests/Common/Cache/CacheTest.php:450
/usr/local/share/poser/vendor/phpunit/phpunit/src/TextUI/Command.php:149
/usr/local/share/poser/vendor/phpunit/phpunit/src/TextUI/Command.php:100

--

There were 2 failures:

1) Doctrine\Tests\Common\Cache\ApcCacheTest::testFlushAll
Failed asserting that false is true.

cache/tests/Doctrine/Tests/Common/Cache/CacheTest.php:346
/usr/local/share/poser/vendor/phpunit/phpunit/src/TextUI/Command.php:149
/usr/local/share/poser/vendor/phpunit/phpunit/src/TextUI/Command.php:100

2) Doctrine\Tests\Common\Cache\ApcCacheTest::testFlushAllAndNamespaceVersioningBetweenCaches
Failed asserting that false is true.

cache/tests/Doctrine/Tests/Common/Cache/CacheTest.php:382
/usr/local/share/poser/vendor/phpunit/phpunit/src/TextUI/Command.php:149
/usr/local/share/poser/vendor/phpunit/phpunit/src/TextUI/Command.php:100
```

flack:
Travis failure seems unrelated (looks like it couldn't install some extensions). The first Scrutinizer issue seems like a false positive (`deleteMulti` exists since 2.0 released in 2012), the second one (missing visibility declaration in `deleteMultiple`) is true, but this is in line with the other `Multi` interfaces. i could add a commit to address all three, or send a separate PR if you find it necessary

flack:
As for supported versions: 
- `memcached` >= 2.0.0 from 2012 
- `apcu` >= 4.0.0 from (seems to have been there from the beginning, at least it's already present in the intial import to github)
- `redis` support was added in Nov 2009, so it might have been in 1.2.0 released in Mar 2010 or in 2.0.0 released in Sep 2010. PECL lists the first public release as 2.2.3, so it would have definitely been in there

lib/Doctrine/Common/Cache/MemcachedCache.php
```diff
@@ -92,6 +92,16 @@ protected function doSaveMultiple(array $keysAndValues, $lifetime = 0)
     /**
      * {@inheritdoc}
      */
+    protected function doDeleteMultiple(array $keys)
+    {
+        $stat = $this->memcached->deleteMulti($keys);
+        return count(array_unique($stat) > 1)
+            || end($stat) === true;
```
Ocramius: Is memcached really this ugly? :|

lib/Doctrine/Common/Cache/CacheProvider.php
```diff
@@ -286,6 +297,26 @@ protected function doSaveMultiple(array $keysAndValues, $lifetime = 0)
...
+     *
+     * @return bool TRUE if the operation was successful, FALSE if it wasn't.
+     */
+    protected function doDeleteMultiple(array $keys)
+    {
+        $success = true;
+
+        foreach ($keys as $key) {
+            if (!$this->doDelete($key)) {
+                $success = false;
```
Ocramius: I know that this may sound like an ugly question, but is this scenario covered by the tests? (doesn't seem like it)

lib/Doctrine/Common/Cache/CacheProvider.php
```diff
@@ -119,6 +120,16 @@ public function saveMultiple(array $keysAndValues, $lifetime = 0)
     /**
      * {@inheritdoc}
      */
+    public function deleteMultiple(array $keys)
+    {
+        $namespacedKeys = array_map([$this, 'getNamespacedId'], $keys);
```
Ocramius: This assignment seems redundant

lib/Doctrine/Common/Cache/MemcachedCache.php
Ocramius: @flack what about `count(array_filter($stat)) === count($keys)`?

Ocramius: Needs testing against:

 * empty set of keys
 * one key (success)
 * one key (failure)
 * multiple keys (success)
 * multiple keys (a failure)
 * multiple keys (all failures)

lib/Doctrine/Common/Cache/CacheProvider.php
Ocramius: Correct

lib/Doctrine/Common/Cache/MemcachedCache.php
flack: Yes, that would require all operations to be successful. Like I've said, it can easily be changed. I just wasn't sure what the preferred behavior would be. 

For example, if you look at `RedisCache`, the normal `doDelete` call returns `true` even if no entry was affected. The logic you propose would have to return `false` in such a case, causing an inconsistency between drivers. On the other hand, driver behavior doesn't look that consistent to begin with afaict, so I'll take whichever version you prefer :-)

lib/Doctrine/Common/Cache/CacheProvider.php
flack: Let's put it this way: I've based my test on the `saveMultiple` test. So it seems that neither one of them is test the failure case :-)

lib/Doctrine/Common/Cache/MemcachedCache.php
Ocramius: @flack yeah, we can't guarantee consistency across all drivers (unless we check for existence and then delete), but when the driver may fail, then we should at least test that scenario. Mocking out memcached is a way to achieve this in this scenario, for example (doesn't need to be an actual functional test)

Ocramius: Test would be in the memcached-specific test class

lib/Doctrine/Common/Cache/CacheProvider.php
Ocramius: Yeah, probably overlooked at that time. We should still strive for adding coverage here

Ocramius:
Bleh, I just saw #155 is a dupe of this... :S

flack:
@Ocramius so how do you want to proceed? Do you prefer to work on #155 or this one? For me, either way is fine, I just want the functionality :-)

Ocramius:
@flack not sure yet - possibly worth integrating the tests from #155 in here (merge from that branch?)

lib/Doctrine/Common/Cache/MultiDeleteCache.php
```diff
@@ -0,0 +1,38 @@
...
+ */
+interface MultiDeleteCache
+{
+    /**
+     * Returns a boolean value indicating if the operation succeeded.
+     *
+     * @param string[] $keys Array of keys to delete from cache
+     * @return bool TRUE if the operation was successful, FALSE if it wasn't.
+     */
+    function deleteMultiple(array $keys);
```
malukenho: It's missing visibility

lcobucci:
Almost everything of this PR was merged on https://github.com/doctrine/cache/pull/155#issuecomment-267788928.

@Ocramius we just need to get the APCu and Redis implementations for the `doDeleteMultiple`(and test them)

PR #191, opened 2016-10-27
add doFetchMultiple method for ChainCache

moknomo:
ChainCache kill perfomance with providers accepting fetchMultiple. For example in my case I use chain ArrayAdapter-RedisAdapter and redis did "get" for each key. For 2000 keys it about 0.18 sec but with one "mGet" time about 0.01 sec!

### DIFF
--- a/lib/Doctrine/Common/Cache/ChainCache.php
+++ b/lib/Doctrine/Common/Cache/ChainCache.php
@@ -52,7 +52,34 @@ class ChainCache extends CacheProvider
             $cacheProvider->setNamespace($namespace);
         }
     }
+    
+    /**
+     * {@inheritDoc}
+     */
+    protected function doFetchMultiple(array $keys)
+    {
+        $returnValues = [];
 
+        $keys = array_combine($keys, $keys);
+        foreach ($this->cacheProviders as $idProvider => $cacheProvider) {
+            $items = $cacheProvider->doFetchMultiple($keys);
+            foreach ($items as $key => $value) {
+                if(false !== $value) {
+                    $returnValues[$key] = $value;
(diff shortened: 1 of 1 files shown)


lib/Doctrine/Common/Cache/ChainCache.php
```diff
@@ -52,7 +52,34 @@ public function setNamespace($namespace)
...
+    {
+        $returnValues = [];
+
+        $keys = array_combine($keys, $keys);
+        foreach ($this->cacheProviders as $idProvider => $cacheProvider) {
+            $items = $cacheProvider->doFetchMultiple($keys);
+            foreach ($items as $key => $value) {
+                if(false !== $value) {
+                    $returnValues[$key] = $value;
+                    for ($subIDProvider = $idProvider - 1 ; $subIDProvider >= 0 ; $subIDProvider--) {
```
Ocramius: This code is really hard to read :-\

lib/Doctrine/Common/Cache/ChainCache.php
```diff
@@ -52,7 +52,34 @@ public function setNamespace($namespace)
             $cacheProvider->setNamespace($namespace);
         }
     }
+    
+    /**
+     * {@inheritDoc}
+     */
+    protected function doFetchMultiple(array $keys)
```
Ocramius: Can't you simply defer to `fetchMultiple` to sub-providers, if available and the interface supports that?

Ocramius:
Closing as invalid. The correct approach is to:

 * iterate through all adapters
 * use multi-get and multi-set where available in the internal adapter (check via `instanceof`)

Please re-open a new PR with that approach, if you can still work on it. Remember: tests are necessary, so please verify that there is at least code path coverage for any newly introduced code.

PR #196, opened 2016-11-03
Fix temp file rights on shared storage

mikeSimonson:
It is apparently necessary to set the rights of the file both
before and after writing it. Depending if you write the file on a
shared storage or not.


mikeSimonson:
related to this thread.
https://github.com/doctrine/annotations/pull/100

Ocramius:
LGTM. Don't like the fact that there are no tests, but I'm not going to set up a shared storage for this edge case...

PR #197, opened 2016-11-14
Tag support implementation for fetching and invalidate cache entries marked by tags

Smart-Core:
Very useful things.

### DIFF
--- a/lib/Doctrine/Common/Cache/Cache.php
+++ b/lib/Doctrine/Common/Cache/Cache.php
@@ -73,10 +73,11 @@ interface Cache
      * @param int    $lifeTime The lifetime in number of seconds for this cache entry.
      *                         If zero (the default), the entry never expires (although it may be deleted from the cache
      *                         to make place for other entries).
+     * @param array $tags
      *
      * @return bool TRUE if the entry was successfully stored in the cache, FALSE otherwise.
      */
-    public function save($id, $data, $lifeTime = 0);
+    public function save($id, $data, $lifeTime = 0, array $tags = []);
 
     /**
      * Deletes a cache entry.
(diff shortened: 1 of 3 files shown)


lib/Doctrine/Common/Cache/Cache.php
```diff
@@ -73,10 +73,11 @@ public function contains($id);
      * @param int    $lifeTime The lifetime in number of seconds for this cache entry.
      *                         If zero (the default), the entry never expires (although it may be deleted from the cache
      *                         to make place for other entries).
+     * @param array $tags
      *
      * @return bool TRUE if the entry was successfully stored in the cache, FALSE otherwise.
      */
-    public function save($id, $data, $lifeTime = 0);
+    public function save($id, $data, $lifeTime = 0, array $tags = []);
```
Ocramius: This is an interface change, and a BC break. Not mergeable, sorry.

Koc:
@Smart-Core have you looked at Symfony 3.2 PSR6 implementation? Have you considered switching to it? Also Symfony provide more efficient method of invalidation by tag.

Ocramius:
Looking at this again, I don't really see a need for it inside a generic lib like doctrine/cache.

This is a very specific use-case that has very very few consumers, generally. A namespaced cache key reduces the use-case scenario massively too.

Closing, therefore.

@Koc PSR-6 got downvoted by doctrine because of it's useless complexity, but doesn't include tagging AFAIK.

Smart-Core:
it's a pity

Ocramius:
@Smart-Core always consider the 80/20 here: supporting all use-cases doesn't bring much, whereas a custom cache implementation can support them with really not that much coding (see this PR) after all.

This is more like a 5% scenario, so this library is not a good fit for the feature.

PR #198, opened 2016-12-03
Disable PHPUnit feature @runInSeparateProcess 

snsanich:
Hi.

PHPUnit feature @runInSeparateProcess cannot work properly if some extensions are not available.

Step by step:
* include tests/travis/php.ini into php config
* run php7.0 `php --interactive`, you will see error/errors:

`Warning: PHP Startup: Unable to load dynamic library '...php-7.0.13/lib/php/extensions/no-debug-non-zts-20151012/mongo.so' - dlopen(...php-7.0.13/lib/php/extensions/no-debug-non-zts-20151012/mongo.so, 9): image not found in Unknown on line 0`

* now run phpunit - test set is failed
* if you remove @runInSeparateProcess  - pass!

### DIFF
--- a/tests/Doctrine/Tests/Common/Cache/FileCacheTest.php
+++ b/tests/Doctrine/Tests/Common/Cache/FileCacheTest.php
@@ -219,7 +219,6 @@ class FileCacheTest extends \Doctrine\Tests\DoctrineTestCase
     }
 
     /**
-     * @runInSeparateProcess
      * @dataProvider getPathLengthsToTest
      *
      * @covers \Doctrine\Common\Cache\FileCache::getFilename


GrahamCampbell:
> PHPUnit feature @runInSeparateProcess cannot work properly if some extensions are not available

I don't think that's what that error means. That's PHP saying you have a problem with your config file, rather than being related to the testsuite. if you fix you fix your configuration files, then the tests should work correctly.

> if you remove @runInSeparateProcess - pass!

Without knowing much about this project, there surely must have been a good reason why it's running in a separate process?

snsanich:
> Without knowing much about this project, there surely must have been a good reason why it's running in a separate process?

No reasons to use a separate process were found.

> I don't think that's what that error means. That's PHP saying you have a problem with your config file, rather than being related to the testsuite.

Currently I don't know how to implement easy fix for travis environment (php7. php7.1). I think it is affordable price to test project in the most easy way.

Ocramius:
Indeed, this test does not deal with concurrency, so running it in a separate process is not needed.

The fact that a dynamic library can't be loaded is a bug in your own configuration though, and unrelated with this particular change.

Merging anyway, since the change is valid :-)

Issue #161, opened 2016-06-10
Memcached: doContains return true pretty much everytime if the stored value is "false"


riquito:
The current implementation of [MemcachedCache::doContains](https://github.com/doctrine/cache/blob/4e3b8b9/lib/Doctrine/Common/Cache/MemcachedCache.php#L98) return `true` if the stored value is `false` and there is any kind of error code different from RES_NOTFOUND. 

If the error code is something like RES_NOTSTORED, RES_SOME_ERRORS or many others then `doContains` should return `false`.

Tobion:
According to documentation, those return values cannot be returned from a simple get. So please provide a test case to prove something else.

deeky666:
@Tobion unfortunately the docs are not revealing every implementation detail. [There are even tests that cover those constants](https://github.com/php-memcached-dev/php-memcached/search?utf8=%E2%9C%93&q=RES_SOME_ERRORS).

Tobion:
I don't see a test for `get` and `RES_NOTFOUND`.

riquito:
I can't provide a test, but I suppose that if you try a `get` after shutting down memcached you won't get RES_NOTFOUND (can't try right now). At present I'm getting in a production server result code 47, MEMCACHED_SERVER_TEMPORARILY_DISABLED, which causes `doContains()` to return `true` (don't know how to reproduce it on purpose).

deeky666:
@Tobion https://github.com/php-memcached-dev/php-memcached/blob/607860c656a3cea22fb164c4eb71d36ec869c598/tests/experimental/getmulti_partial_error.phpt#L23

Tobion:
That is a multiget. I already said twice that this ticket is about a simple get returning those states. Please read more carefully.

riquito:
Here are some links to the implementation of `get` in the php memcached extension. 

[...]

PR #199, opened 2016-12-04
Memcached : fix doContains return true even if status code is an error (related to #181)

JonasHaouzi:
I couldn't push on the repo. of @omouren so I created a new fork.
Here's the code provided by @omouren with a fixed bug, and tests.

Can you review @Ocramius ?

Thanks.

PR #201, opened 2016-12-06
Fix contains check on memcached.

SpacePossum:
closes https://github.com/doctrine/cache/issues/161
replaces https://github.com/doctrine/cache/pull/181

### DIFF
--- a/lib/Doctrine/Common/Cache/MemcachedCache.php
+++ b/lib/Doctrine/Common/Cache/MemcachedCache.php
@@ -94,10 +94,8 @@ class MemcachedCache extends CacheProvider
      */
     protected function doContains($id)
     {
-        return
-            false === $this->memcached->get($id)
-            ? Memcached::RES_SUCCESS === $this->memcached->getResultCode() // returned is false, which mean either does not exits, failure or false was stored at key
-            : true // returned is not false, so cache contains the item
+        return false !== $this->memcached->get($id)
+            || Memcached::RES_SUCCESS === $this->memcached->getResultCode() // returned is false, which mean either does not exits, failure or `false` as value was stored at key
         ;
     }
 
--- a/lib/Doctrine/Common/Cache/MemcachedCache.php
+++ b/lib/Doctrine/Common/Cache/MemcachedCache.php
@@ -94,8 +94,11 @@ class MemcachedCache extends CacheProvider
      */
     protected function doContains($id)
     {
-        return false !== $this->memcached->get($id)
-            || $this->memcached->getResultCode() !== Memcached::RES_NOTFOUND;
+        return
+            false === $this->memcached->get($id)
+            ? Memcached::RES_SUCCESS === $this->memcached->getResultCode() // returned is false, which mean either does not exits, failure or false was stored at key
+            : true // returned is not false, so cache contains the item
+        ;
     }
 
     /**
--- a/tests/Doctrine/Tests/Common/Cache/MemcachedCacheTest.php
+++ b/tests/Doctrine/Tests/Common/Cache/MemcachedCacheTest.php
@@ -49,6 +49,33 @@ class MemcachedCacheTest extends CacheTest
         $this->assertInstanceOf('Memcached', $this->_getCacheDriver()->getMemcached());
     }
 
+    public function testDoContains()
+    {
+        $driver = $this->_getCacheDriver();
+        $reflection = new \ReflectionClass($driver);
+        $method = $reflection->getMethod('doContains');
+        $method->setAccessible(true);
+
+        $testKey = __CLASS__.'#'.__METHOD__;
+
+        $this->memcached->set($testKey, false);
+
+        $this->assertTrue($method->invokeArgs($driver, [$testKey]));
+        $this->assertFalse($method->invokeArgs($driver, [$testKey.'2']));
+
+        $memcached = new Memcached();
+        $memcached->addServer('0.0.0.1', 11211); // fake server is not available
+
+        $driver = new MemcachedCache();
+        $driver->setMemcached($memcached);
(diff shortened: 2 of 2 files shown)


Ocramius:
Seen #199?

SpacePossum:
>Seen #199?

no, d'oh I thought 161 and 181 was all...
well now you can choose ^_^
(or merge some stuff together, my tests cover more cases (i.e. proof the issue))

ksungcaya:
Any update when this will be merged? Currently waiting for this fix. :D

SpacePossum:
ping @Ocramius any updates you want on this PR?

Ocramius:
@SpacePossum simply didn't get to do any doctrine/cache recently, sorry.

lib/Doctrine/Common/Cache/MemcachedCache.php
```diff
@@ -94,8 +94,11 @@ protected function doSaveMultiple(array $keysAndValues, $lifetime = 0)
      */
     protected function doContains($id)
     {
-        return false !== $this->memcached->get($id)
-            || $this->memcached->getResultCode() !== Memcached::RES_NOTFOUND;
+        return
+            false === $this->memcached->get($id)
+            ? Memcached::RES_SUCCESS === $this->memcached->getResultCode() // returned is false, which mean either does not exits, failure or false was stored at key
```
stof: I would keep the previous ``||``, but changing the result code comparison

tests/Doctrine/Tests/Common/Cache/MemcachedCacheTest.php
```diff
@@ -49,6 +49,33 @@ public function testGetMemcachedReturnsInstanceOfMemcached()
         $this->assertInstanceOf('Memcached', $this->_getCacheDriver()->getMemcached());
     }
 
+    public function testDoContains()
+    {
+        $driver = $this->_getCacheDriver();
+        $reflection = new \ReflectionClass($driver);
+        $method = $reflection->getMethod('doContains');
```
stof: IMO, it would be better to write the test using the public API directly (i.e. ``contains``)

tests/Doctrine/Tests/Common/Cache/MemcachedCacheTest.php
```diff
@@ -49,6 +49,33 @@ public function testGetMemcachedReturnsInstanceOfMemcached()
...
+
+        $testKey = __CLASS__.'#'.__METHOD__;
+
+        $this->memcached->set($testKey, false);
+
+        $this->assertTrue($method->invokeArgs($driver, [$testKey]));
+        $this->assertFalse($method->invokeArgs($driver, [$testKey.'2']));
+
+        $memcached = new Memcached();
+        $memcached->addServer('0.0.0.1', 11211); // fake server is not available
```
stof: you should write 2 separate tests (one covering the storage of ``false`` and another one covering the handling of errors with a fake server) instead of mixing both in the same test. We need them to be able to fail independently. Thus, it would give more info about what each test is covering (as they would each have a descriptive name)

tests/Doctrine/Tests/Common/Cache/MemcachedCacheTest.php
```diff
@@ -51,9 +51,8 @@ public function testGetMemcachedReturnsInstanceOfMemcached()
 
     public function testContainsWithKeyWithFalseAsValue()
     {
-        $testKey = __CLASS__.'#'.__METHOD__;
+        $testKey = __METHOD__;
         $driver = $this->_getCacheDriver();
-
         $this->memcached->set($testKey, false);
```
SpacePossum: I prop. need to fix the NS stuff now I no longer directly test the cache implementation but the public API layer (as requested), thanks for looking!

tests/Doctrine/Tests/Common/Cache/MemcachedCacheTest.php
```diff
@@ -49,6 +49,31 @@ public function testGetMemcachedReturnsInstanceOfMemcached()
...
 
+    public function testContainsWithKeyWithFalseAsValue()
+    {
+        $testKey = __METHOD__;
+        $driver = $this->_getCacheDriver();
+        $reflection = new \ReflectionClass($driver);
+        $method = $reflection->getMethod('getNamespacedId');
+        $method->setAccessible(true);
+        $testKeyNS = $method->invokeArgs($driver, [$testKey]);
+        $this->memcached->set($testKeyNS, false);
```
stof: shouldn't you set the value through doctrine/cache ?

SpacePossum: This will test the code works even if the item was cached in a previous run of the code and not only during a single executing of it. I you think it holds no value I can refactor. 

TBH we are no longer using the caching layer of Doctrine at all (we use a custom one because of other issues we have with it) so I don't want to spend much more time/effort on this PR. Please let me know if you see more issue with this so I can address those in one go.

SpacePossum:
ping @Ocramius let me know if there is anything left you want me to do :)

Ocramius:
@SpacePossum looks good to me, but won't get to merge+backport these days, as I'm completely focused on client work.

SpacePossum:
@Ocramius thanks for letting me know :) good luck with the client work!

SpacePossum:
I'm closing my PR's I lost interest in as I do not want to leave the expression that I'll follow up on those.
Thanks to all the reviewers!

Ocramius:
Re-opened - no need to follow up, @SpacePossum: if you can't help out later, just unsubscribe from this thread. We know it's up to us now.

Ocramius:
Manually merged via doctrine/cache@767356d1f4ca7170fb7e9a701ff9755ebbb9736b

PR #203, opened 2017-01-02
Removes support for PHP 5.x

railto:
Removes support for PHP 5.x from composer.json and .travis.yml

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,6 @@ cache:
         - $HOME/.composer/cache
 
 php:
-    - 5.6
     - 7.0
     - 7.1
     - hhvm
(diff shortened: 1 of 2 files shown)


bordeux:
why?

Ocramius:
@bordeux PHP 5.x is now in security-only mode as per http://php.net/supported-versions.php

Older versions of the cache component will still be patched for security issues, but we're not maintaining active support for unsupported PHP versions.

railto:
Closing as I do not have the ability to get this to work with HHVM tests

Ocramius:
Reopening: still needs to happen, even if you don't work on the PR yourself.

lcobucci:
Handled in #219 and #220

PR #211, opened 2017-02-15
Do not force igbinary serializer for redis if the module is installed.

### LINKED ISSUE #150: Impossible to override Redis serializer

mvanbaak:
This fixes https://github.com/doctrine/cache/issues/150

Also, the RedisCache is now inline with MemcachedCache in that it uses the serializer specified (or auto-set) on the passed Cache object.

### DIFF
--- a/tests/Doctrine/Tests/Common/Cache/RedisCacheTest.php
+++ b/tests/Doctrine/Tests/Common/Cache/RedisCacheTest.php
@@ -35,18 +35,6 @@ class RedisCacheTest extends CacheTest
         $this->assertInstanceOf('Redis', $this->_getCacheDriver()->getRedis());
     }
 
-    public function testSerializerOptionWithOutIgbinaryExtension()
-    {
-        if (defined('Redis::SERIALIZER_IGBINARY') && extension_loaded('igbinary')) {
-            $this->markTestSkipped('Extension igbinary is loaded.');
-        }
-
-        $this->assertEquals(
-            \Redis::SERIALIZER_PHP,
-            $this->_getCacheDriver()->getRedis()->getOption(\Redis::OPT_SERIALIZER)
-        );
-    }
-
     /**
      * {@inheritDoc}
      */
--- a/tests/Doctrine/Tests/Common/Cache/RedisCacheTest.php
+++ b/tests/Doctrine/Tests/Common/Cache/RedisCacheTest.php
@@ -35,6 +35,18 @@ class RedisCacheTest extends CacheTest
         $this->assertInstanceOf('Redis', $this->_getCacheDriver()->getRedis());
     }
 
+    public function testSerializerOptionWithOutIgbinaryExtension()
+    {
+        if (defined('Redis::SERIALIZER_IGBINARY') && extension_loaded('igbinary')) {
+            $this->markTestSkipped('Extension igbinary is loaded.');
+        }
+
+        $this->assertEquals(
+            \Redis::SERIALIZER_PHP,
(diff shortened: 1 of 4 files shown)


tests/Doctrine/Tests/Common/Cache/RedisCacheTest.php
```diff
@@ -35,18 +35,6 @@ public function testGetRedisReturnsInstanceOfRedis()
         $this->assertInstanceOf('Redis', $this->_getCacheDriver()->getRedis());
     }
 
-    public function testSerializerOptionWithOutIgbinaryExtension()
```
Ocramius: This is a BC break for those using igbinary

Ocramius (changes requested):

tests/Doctrine/Tests/Common/Cache/RedisCacheTest.php
mvanbaak: yes, but unfortunately it's the only way to allow people to choose a serializer different from igbinary when the igbinary module is installed in their environment (which breaks a lot of things on PHP-7 and CommandLine scripts)
If you have a better way to do it I'm all ears. I already thought of adding an extra parameter but that will make things a lot more ugly, and will make the RedisCache even more different from the other implementations.

Ocramius: Then a new implementation is needed (different class name)

mvanbaak: Ok, I can do that :) Any ideas? I'm very bad at that. RedisWithPresetSerializerCache ?

Ocramius: Try something, then we review? :-P

In general, yes, a new symbol is a better solution than a BC break, especially on such a widespread library.

apflieger:
Any news? I can't wait to have that :)

lcobucci:
@apflieger as per @Ocramius the approach to solve this issue would be having a new symbol instead of breaking BC.
The author of the PR - probably - didn't have time to address this however you're quite welcome to send us another PR so that we can review it 😉

mvanbaak:
Sorry all, it took some time.

Didn't have time to do this earlier (and we were using my fork in production so the urgent need was no longer there)

ishuowang:
In my project with RedisCache, because forced igbinary serializer caused some problems, then I replaced with PredisCache. I think in the creation of the client can provide an optional parameter to determine the OPT_SERIALIZER way.
Some code:
public function setRedis(Redis $redis, $opt_serializer = false)
    {
        $opt_serializer = $opt_serializer?:$this->getSerializerValue();
        $redis->setOption(Redis::OPT_SERIALIZER, $opt_serializer );
        $this->redis = $redis;
    }

lcobucci:
@mvanbaak no worries, we're all pretty busy 😄 

Can you check these failing tests?

yellow1912:
Any news on this? Seem like the class name is already changed as advised by @Ocramius

alcaeus:
> Any news on this?

No, still waiting on test fixes from @mvanbaak as asked by @lcobucci.

alcaeus:

PR #214, opened 2017-03-06
Combine multi operation interfaces into a single one

lcobucci:
Since we always depend on the three interfaces there's no reason to segregate them.

And that's also important to be able to implement https://github.com/doctrine/doctrine2/issues/6241


guilhermeblanco:
:+1:

PR #215, opened 2017-03-06
Fix off-by-one accessing last object in list

robocoder:


Ocramius:
@robocoder needs tests

robocoder:
Oh, it isn't obvious that `$objectList[count($objectList)];` is NULL?  ;)

Ocramius:
Can be as clear as the sky, but needs to be tested to prevent regressions.

robocoder:
The unit test I committed is broken. I'll troubleshoot it later (tonight?).

robocoder:
:taxa:

robocoder:
The Travis failures are with php7.2 and RedisCache (not Riak).

robocoder:
Rebased with master.

Ocramius:
@guilhermeblanco I suppose this needs to be backported into the current release branch too...

PR #217, opened 2017-04-30
Redis: Remove useless ifdef for HHVM

ostrolucky:
That condition is no longer needed, it was added in past to fix tests when extension_loaded('igbinary') piece didn't exist yet

### DIFF
--- a/lib/Doctrine/Common/Cache/RedisCache.php
+++ b/lib/Doctrine/Common/Cache/RedisCache.php
@@ -175,10 +175,6 @@ class RedisCache extends CacheProvider
      */
     protected function getSerializerValue()
     {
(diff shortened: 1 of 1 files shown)


Ocramius:
@ostrolucky is this to be marked `invalid`? What's the reasoning for closing, for future reference?

ostrolucky:
Lot of tests failed and I assumed it's because of my change. Didn't investigate further if it's really because of that, though. Maybe we should, so ok, reopening.

ostrolucky:
Ok I tried it again and travis fails without these changes too, so this change is probably ok.

lcobucci:
Handled in #219

PR #218, opened 2017-05-01
Run APCu and APC on PHP 7.x too

lcobucci:
APC(u) tests are being skipped for PHP 7.x and they shouldn't (IMO)


lcobucci:
Still not running APCu tests, will check this later

lcobucci:
Alright, I give up... this is not looking good to me at all 😄 
Perhaps we should just let some things (like `ApcCache`) die (as in `doctrine/cache` v2).

Ocramius:
@krakjoe would you suggest to keep APCu compatibility in the future?

PR #219, opened 2017-05-18
Drop HHVM support

lcobucci:
Starting HHVM support removal on our repos, related to doctrine/doctrine2#6424

### DIFF
--- a/lib/Doctrine/Common/Cache/RedisCache.php
+++ b/lib/Doctrine/Common/Cache/RedisCache.php
@@ -175,10 +175,6 @@ class RedisCache extends CacheProvider
      */
     protected function getSerializerValue()
     {
-        if (defined('HHVM_VERSION')) {
-            return Redis::SERIALIZER_PHP;
-        }
-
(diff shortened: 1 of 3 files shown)


lib/Doctrine/Common/Cache/RedisCache.php
```diff
@@ -175,10 +175,6 @@ protected function doGetStats()
      */
     protected function getSerializerValue()
     {
-        if (defined('HHVM_VERSION')) {
```
Ocramius: This still counts as a BC break. While we are indeed dropping HHVM support, this doesn't mean that we should also break code for them ;-)

lcobucci: This is the exact same change as in https://github.com/doctrine/cache/pull/217 and it's useless since `extension_loaded('igbinary')` is checked

Ocramius: Gotcha! :+1:

Issue #142, opened 2016-02-11
Support for MongoDBCache on PHP 7 using ext-mongodb


holtkamp:
Related to https://github.com/doctrine/mongodb/issues/239: in order to currently use [MongoDBCache](https://github.com/doctrine/cache/blob/master/lib/Doctrine/Common/Cache/MongoDBCache.php) `ext-mongo` is required. This extension has been replaced by [ext-mongodb](https://pecl.php.net/package/mongodb) on PHP 7.

Effectively this means that MongoDBCache can currently not be used on PHP 7.

holtkamp:
Elaborating on this, what would be the most viable approach to support PHP 7?

What I can think of

1. create new class, for example "MongoDBCacheNG" (Next Generation ;)) which depends on ext-mongodb
- typecheck depending on which extension is loaded (ext-mongo / ext-mongodb) on either:
 - "MongoCollection" / "MongoDB\Collection"
 - "MongoDate" / "MongoDB\BSON\UTCDateTime"
 - "MongoBinData" / "MongoDB\BSON\Binary"
 - "MongoCursorException" / "MongoDB\Driver\Exception\Exception"
- new major version of Doctrine\Cache
 - only types as used in ext-mongodb and depend on PHP 7 only

I think the first approach would be the simplest in terms of development, testing and maintenance...

Ocramius:
Handled in #222

PR #220, opened 2017-05-20
Require PHP 7.1

lcobucci:
Related to doctrine/doctrine2#6425


tests/Doctrine/Tests/Common/Cache/MongoDBCacheTest.php
```diff
@@ -6,20 +6,34 @@
...
+
     /**
      * @var MongoCollection
      */
     private $collection;
 
     protected function setUp()
     {
-        $mongo = new MongoClient();
+        if (static::$skip === true) {
```
lcobucci: This is nasty, I known... the idea is to speed up the failure.

@Ocramius @alcaeus I'd love if you could provide a better approach.

Ocramius: Use a `@requires` annotation at class level?

Ocramius: Also: speed is not a problem on this project: if you want to work on that, we have some tests using `sleep()` 😅

lcobucci: `sleep()`??? 😭 

F*** that then, the `@requires` will not work because the connection is "lazy" so it fails only when trying to do something with it.

I'll keep the try/catch and remove that nasty static property... thanks

composer.json
```diff
@@ -13,12 +13,16 @@
...
-        "php": "~5.6|~7.0"
+        "php": "~7.1"
     },
     "require-dev": {
+        "alcaeus/mongo-php-adapter": "^1.1",
         "phpunit/phpunit": "^5.7",
         "predis/predis":   "~1.0"
     },
+    "suggest": {
+        "alcaeus/mongo-php-adapter": "Required to use MongoDB driver"
```
alcaeus: Just cross-referencing #142. Rewriting `MongoDBCache` would be a BC break, so the suggestion is good, but it's not a long-term solution. I'll create a pull request to deprecate the current class, we'll just have to think of some clever naming for the new cache class that uses `ext-mongodb`.

lcobucci: @alcaeus indeed, I decided to leave it for a different PR to keep things consistent. Thanks for taking it over ;)

PR #221, opened 2017-05-26
Create expiration index when getting expired entry

### LINKED ISSUE #139: AutoExpiration index of MongoCollections not configured automatically

alcaeus:
Fixes #139, Supersedes #140.

Instead of always creating the index on class instantiation, this solution only creates the index once per process and only if receiving an expired cache entry (indicating that the index hasn't been properly created). Overrides the lifetime test case to test index creation. My machine didn't quite remove expired entries exactly at their expiration time, so I excluded that from the test case.

### DIFF
--- a/tests/Doctrine/Tests/Common/Cache/MongoDBCacheTest.php
+++ b/tests/Doctrine/Tests/Common/Cache/MongoDBCacheTest.php
@@ -68,6 +68,17 @@ class MongoDBCacheTest extends CacheTest
         self::assertFalse($cache->save('foo', 'bar'));
     }
 
+    public function testLifetime() : void
+    {
+        $cache = $this->_getCacheDriver();
+        $cache->save('expire', 'value', 1);
+        $this->assertCount(1, $this->collection->getIndexInfo());
+        $this->assertTrue($cache->contains('expire'), 'Data should not be expired yet');
+        sleep(2);
+        $this->assertFalse($cache->contains('expire'), 'Data should be expired');
+        $this->assertCount(2, $this->collection->getIndexInfo());
+    }
+
     protected function _getCacheDriver() : CacheProvider
     {
         return new MongoDBCache($this->collection);
--- a/lib/Doctrine/Common/Cache/MongoDBCache.php
+++ b/lib/Doctrine/Common/Cache/MongoDBCache.php
@@ -60,6 +60,11 @@ class MongoDBCache extends CacheProvider
     private $collection;
 
     /**
+     * @var bool
+     */
+    private $expirationIndexCreated = false;
+
+    /**
      * Constructor.
      *
      * This provider will default to the write concern and read preference
@@ -89,6 +94,7 @@ class MongoDBCache extends CacheProvider
         }
 
         if ($this->isExpired($document)) {
+            $this->createExpirationIndex();
             $this->doDelete($id);
             return false;
         }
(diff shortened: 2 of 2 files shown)


tests/Doctrine/Tests/Common/Cache/MongoDBCacheTest.php
```diff
@@ -68,6 +68,22 @@ public function testMongoCursorExceptionsDoNotBubbleUp() : void
         self::assertFalse($cache->save('foo', 'bar'));
     }
 
+    public function testLifetime() : void
+    {
+        $cache = $this->_getCacheDriver();
+        $cache->save('expire', 'value', 1);
+        $this->assertCount(1, $this->collection->getIndexInfo());
+        $this->assertTrue($cache->contains('expire'), 'Data should not be expired yet');
+        sleep(2);
```
Ocramius: Don't like these, but hey, integration testing! 🤷

tests/Doctrine/Tests/Common/Cache/MongoDBCacheTest.php
```diff
@@ -68,6 +68,22 @@ public function testMongoCursorExceptionsDoNotBubbleUp() : void
...
+        $this->assertFalse($cache->contains('expire'), 'Data should be expired');
+        $this->assertCount(2, $this->collection->getIndexInfo());
+
+        $cache->save('expire', 'value', 1);
```
Ocramius: What's the reason for duplicating this block?

tests/Doctrine/Tests/Common/Cache/MongoDBCacheTest.php
alcaeus: Me neither, but yay! \o/

tests/Doctrine/Tests/Common/Cache/MongoDBCacheTest.php
alcaeus: Was left over from previous tests. Removed.

PR #222, opened 2017-05-26
Add cache provider for ext-mongodb

### LINKED ISSUE #142: Support for MongoDBCache on PHP 7 using ext-mongodb

alcaeus:
Fixes #142.

Creates two new internal classes containing the actual logic, one for the legacy driver, one for ext-mongodb. The existing `MongoDBCache` class just proxies all calls to the correct implementation, depending on which collection class was given in the constructor. Since constructor arguments may change in extending classes, this shouldn't cause a BC break. With 2.0, `MongoDBNGCache` will be renamed `MongoDBCache` and the old classes removed to drop support for ext-mongo.

### DIFF
--- a/lib/Doctrine/Common/Cache/LegacyMongoDBCache.php
+++ b/lib/Doctrine/Common/Cache/LegacyMongoDBCache.php
@@ -1,0 +1,194 @@
+<?php
+/*
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ * This software consists of voluntary contributions made by many individuals
+ * and is licensed under the MIT license. For more information, see
+ * <http://www.doctrine-project.org>.
+ */
+
+namespace Doctrine\Common\Cache;
+
+use MongoBinData;
+use MongoCollection;
+use MongoCursorException;
+use MongoDate;
+
+/**
+ * MongoDB cache provider.
+ *
+ * @author Jeremy Mikola <jmikola@gmail.com>
+ * @internal Do not use - will be removed in 2.0. Use MongoDBCache instead
+ */
+class LegacyMongoDBCache extends CacheProvider
+{
+    /**
+     * @var MongoCollection
+     */
+    private $collection;
+
+    /**
+     * @var bool
+     */
+    private $expirationIndexCreated = false;
+
+    /**
+     * Constructor.
+     *
+     * This provider will default to the write concern and read preference
+     * options set on the MongoCollection instance (or inherited from MongoDB or
+     * MongoClient). Using an unacknowledged write concern (< 1) may make the
+     * return values of delete() and save() unreliable. Reading from secondaries
+     * may make contain() and fetch() unreliable.
+     *
+     * @see http://www.php.net/manual/en/mongo.readpreferences.php
+     * @see http://www.php.net/manual/en/mongo.writeconcerns.php
+     * @param MongoCollection $collection
+     */
+    public function __construct(MongoCollection $collection)
+    {
+        @trigger_error('Using the legacy MongoDB cache provider is deprecated and will be removed in 2.0', E_USER_DEPRECATED);
+        $this->collection = $collection;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function doFetch($id)
+    {
+        $document = $this->collection->findOne(['_id' => $id], [MongoDBCache::DATA_FIELD, MongoDBCache::EXPIRATION_FIELD]);
+
+        if ($document === null) {
+            return false;
+        }
+
+        if ($this->isExpired($document)) {
+            $this->createExpirationIndex();
+            $this->doDelete($id);
+            return false;
+        }
+
+        return unserialize($document[MongoDBCache::DATA_FIELD]->bin);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function doContains($id)
+    {
+        $document = $this->collection->findOne(['_id' => $id], [MongoDBCache::EXPIRATION_FIELD]);
+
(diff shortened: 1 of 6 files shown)


lib/Doctrine/Common/Cache/LegacyMongoDBCache.php
```diff
@@ -0,0 +1,176 @@
...
+                    MongoDBCache::EXPIRATION_FIELD => ($lifeTime > 0 ? new MongoDate(time() + $lifeTime) : null),
+                    MongoDBCache::DATA_FIELD => new MongoBinData(serialize($data), MongoBinData::BYTE_ARRAY),
+                ]],
+                ['upsert' => true, 'multiple' => false]
+            );
+        } catch (MongoCursorException $e) {
+            return false;
+        }
+
+        return isset($result['ok']) ? $result['ok'] == 1 : true;
```
Ocramius: Is the loose comparison intentional?

lib/Doctrine/Common/Cache/LegacyMongoDBCache.php
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function doDelete($id)
+    {
+        $result = $this->collection->remove(['_id' => $id]);
+
+        return isset($result['ok']) ? $result['ok'] == 1 : true;
```
Ocramius: Same

lib/Doctrine/Common/Cache/LegacyMongoDBCache.php
```diff
@@ -0,0 +1,176 @@
...
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function doFlush()
+    {
+        // Use remove() in lieu of drop() to maintain any collection indexes
+        $result = $this->collection->remove();
+
+        return isset($result['ok']) ? $result['ok'] == 1 : true;
```
Ocramius: Same

lib/Doctrine/Common/Cache/LegacyMongoDBCache.php
```diff
@@ -0,0 +1,176 @@
...
+            'recordStats' => 0,
+            'repl' => 0,
+        ]);
+
+        $collStats = $this->collection->db->command(['collStats' => 1]);
+
+        return [
+            Cache::STATS_HITS => null,
+            Cache::STATS_MISSES => null,
+            Cache::STATS_UPTIME => (isset($serverStatus['uptime']) ? (int) $serverStatus['uptime'] : null),
```
Ocramius: You can use php7 syntax for all the `isset()` calls in this patch

lib/Doctrine/Common/Cache/MongoDBCache.php
```diff
@@ -55,25 +56,33 @@ class MongoDBCache extends CacheProvider
...
     {
+        if ($collection instanceof MongoCollection) {
+            @trigger_error('Using a MongoCollection instance for creating a cache adapter is deprecated and will be removed in 2.0', E_USER_DEPRECATED);
+            $this->provider = new LegacyMongoDBCache($collection);
+        } elseif ($collection instanceof Collection) {
+            $this->provider = new MongoDBNGCache($collection);
+        } else {
+            throw new \InvalidArgumentException('Invalid collection given - expected a MongoCollection or MongoDB\Collection instance');
+        }
         $this->collection = $collection;
```
Ocramius: Spacing: add an empty line before this one

lib/Doctrine/Common/Cache/MongoDBNGCache.php
```diff
@@ -0,0 +1,221 @@
...
+use MongoDB\Database;
+use MongoDB\Driver\Exception\Exception;
+use MongoDB\Model\BSONDocument;
+
+/**
+ * MongoDB cache provider.
+ *
+ * @internal Do not use - will be removed in 2.0. Use MongoDBCache instead
+ */
+class MongoDBNGCache extends CacheProvider
```
Ocramius: What does `NG` stand for?

lib/Doctrine/Common/Cache/MongoDBNGCache.php
```diff
@@ -0,0 +1,221 @@
...
+        return true;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function doSave($id, $data, $lifeTime = 0)
+    {
+        try {
+            $this->collection->updateOne(
```
Ocramius: Does this call return a success/failure flag?

lib/Doctrine/Common/Cache/MongoDBNGCache.php
```diff
@@ -0,0 +1,221 @@
...
+        return true;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function doDelete($id)
+    {
+        try {
+            $this->collection->deleteOne(['_id' => $id]);
```
Ocramius: Same question

lib/Doctrine/Common/Cache/MongoDBNGCache.php
```diff
@@ -0,0 +1,221 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function doFlush()
+    {
+        try {
+            // Use remove() in lieu of drop() to maintain any collection indexes
+            $this->collection->deleteMany([]);
```
Ocramius: Same as above

lib/Doctrine/Common/Cache/LegacyMongoDBCache.php
alcaeus: This was copied from the old `MongoDBCache` class. But yes, some versions would return an int, others a float. So, loose comparison it is.

lib/Doctrine/Common/Cache/MongoDBNGCache.php
alcaeus: "New generation". Wasn't sure what else to call it. Suggestions are welcome ;)

lib/Doctrine/Common/Cache/MongoDBNGCache.php
alcaeus: It will throw an exception in failure cases, unless we're dealing with unacknowledged writes in which it will always return true. The command returns the number of matched and updated documents, but we're not interested in that number.

lib/Doctrine/Common/Cache/MongoDBNGCache.php
alcaeus: Same answer

lib/Doctrine/Common/Cache/MongoDBNGCache.php
alcaeus: Same as above ;)

lib/Doctrine/Common/Cache/MongoDBNGCache.php
Ocramius: If this is about the php7 compatible one, I'd suggest factoring that in the name.

alcaeus: I'd go with either
* `ExtMongoDBCache` (named for ext-mongodb)
* `Php7MongoDBCache`

lib/Doctrine/Common/Cache/MongoDBNGCache.php
Ocramius: :+1:

lib/Doctrine/Common/Cache/MongoDBNGCache.php
Ocramius: @alcaeus `ExtMongoDBCache` is probably sane

Ocramius:
:ship: nice work! Also, I'll :hocho: you if you squash stuff again mid-review >.<

alcaeus:
> Also, I'll 🔪 you if you squash stuff again mid-review >.<

I'll buy you a beer for your troubles when we're face to face next time. I'll still have somebody check you for knives before doing it.

Ocramius:
I should do this more often if I get treated to :beers:!

alcaeus:
Oh my...what have I done?!?

PR #223, opened 2017-06-13
Redis: Set serialization option

gjuric:
This pull request allows one to explicitly set the serializer that is going to be used by Redis when storing data.

We got hit by this when we upgraded to PHP 7.1 and the Redis extension pulled in igbinary extension as a dependency and all of the existing data stored in Redis was not recognized anymore. Not to mention that we reused the Redis connection for some Redis specific stuff as well and the serializer option was changed in those cases as well :(

I have a pull request for https://github.com/doctrine/DoctrineCacheBundle in the making but it depends on this one.

### DIFF
--- a/lib/Doctrine/Common/Cache/RedisCache.php
+++ b/lib/Doctrine/Common/Cache/RedisCache.php
@@ -36,14 +36,21 @@ class RedisCache extends CacheProvider
     private $redis;
 
     /**
+     * @var int|null
+     */
+    private $serializer;
+
+    /**
      * Sets the redis instance to use.
      *
      * @param Redis $redis
+     * @param int|null $serializer
      *
      * @return void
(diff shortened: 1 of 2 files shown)


Ocramius:
See https://github.com/doctrine/cache/pull/211

Ocramius:
See https://github.com/doctrine/cache/pull/211

gjuric:
That was fast, it looks like my google-fu skills failed me this time.

Buy, If you take a look at my solution you will see that it is not only backwards compatible and includes tests, but it also allows you to set the serializer option if you want to.

I kinda agree that this should probably be set from outside this library, but if this is the only place where one wants to use Redis then it would be handy to have it as an option.

alcaeus:

Issue #224, opened 2017-07-01
Call to undefined function Doctrine\Common\Cache\apcu_fetch()


Laurent3170:
When I clear the cache using `php bin/console doctrine:cache:clear-metadata --env=prod;`or `php bin/console doctrine:cache:clear-query --env=prod;` I receive the error `Call to undefined function Doctrine\Common\Cache\apcu_fetch()`.

The function `apcu_fetch()`is called from https://github.com/doctrine/cache/blob/master/lib/Doctrine/Common/Cache/ApcuCache.php but it doesn't exist in the file, neither in the project...
Do I miss something ?

Ocramius:
It's the `ext-apcu` extension

On 1 Jul 2017 3:56 PM, "Laurent Marquet" <notifications@github.com> wrote:

When I clear the cache using php bin/console doctrine:cache:clear-metadata
--env=prod;or php bin/console doctrine:cache:clear-query --env=prod; I
receive the error Call to undefined function Doctrine\Common\Cache\apcu_
fetch().

The function apcu_fetch()is called from https://github.com/doctrine/
cache/blob/master/lib/Doctrine/Common/Cache/ApcuCache.php but it doesn't
exist in the file, neither in the project...
Do I miss something ?

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/doctrine/cache/issues/224>, or mute the thread

Laurent3170:
Thanks for your answer but can you explain a little bit more what should be done?

lcobucci:
@Laurent3170 you should make sure that the mentioned extension is installed on your environment.

Ocramius:
Closing as `invalid`

Laurent3170:
@Ocramius I can understand that you close this, as it seems not to be an issue directly related to the project, but it's linked with, as the `apcu_fetch()` is called by it.
So it can't work if the external library is not installed, and it may be indicated somewhere (maybe it is but I haven't found).
@lcobucci thanks, I will have a look at it, when I have time. For the time being I can only find information about php-pecl-apcu

Ocramius:
> So it can't work if the external library is not installed, and it may be indicated somewhere (maybe it is but I haven't found).

Yes, feel free to send a patch that verifies if the APCu extension is installed at boot time, but it's not really needed out of the box.

Laurent3170:
[...]

PR #225, opened 2017-07-06
Adds a RunTime Exception for APCu Not Installed

LaurentMarquet:
Prevent user if APCu is not installed #224

### DIFF
--- a/lib/Doctrine/Common/Cache/ApcuCache.php
+++ b/lib/Doctrine/Common/Cache/ApcuCache.php
@@ -29,6 +29,16 @@ namespace Doctrine\Common\Cache;
 class ApcuCache extends CacheProvider
 {
     /**
+     * Throws an error if APCu is not installed
+     */
+    public function __construct()
+    {
+        if (! function_exists('apcu_fetch')) {
+            throw new RuntimeException('APCu is not installed: please install APCu (i.e. "pecl install apcu") before using the APCu cache');
+        }
+    }
+    
+    /**
      * {@inheritdoc}
      */
     protected function doFetch($id)


Ocramius:
Heya!

Can you try adding a test case expecting an exception?

Poke me if you need any guidance

Laurent3170:
Yes I need guidance!

Laurent3170:
Please :-)

Ocramius:
@Laurent3170 you need something similar to https://github.com/doctrine/cache/blob/58f40845d8091b8cb768d6e3a49f685c8dff6928/tests/Doctrine/Tests/Common/Cache/ApcuCacheTest.php#L15, except that it must be a new test that is only skipped when APCu is actually enabled.

Use `$this->expectException(ExceptionType::class)` to verify that an exception is thrown at instantiation time.

Laurent3170:
So... Teacher, here's my homework ;-)
Not really sure that it will work, but I don't know how to test...
```php
public function testApcuExtension() : void
    {
        if (! function_exists('apcu_fetch')) {
            $this->expectException(RuntimeException::ApcuCache);

            new ApcuCache();
        } else {
            $this->markTestSkipped('APCu is actually enabled');
        }
    }
```
I don't know neither if I can add the file `ApcuCacheTest.php` to this PR or if I must create a new one.
Thanks your help

Ocramius:
@Laurent3170 a new test class is needed, as the original has a `$this->markTestSkipped()` call in `setUp()`. Just toss in a new file with a descriptive name :-)

Laurent3170:
So I did another PR #227

Ocramius:
Closing here, moved to #227

Ocramius:
Actually, never mind, my mistake. The tests of #227 should be incorporated in this patch (just push to the same branch). As for the test itself, I'm providing a review on #227 and closing that one instead.

Laurent3170:
I will only be able to look at this in 10-12 days. Thanks for all your time and information!

Laurent3170:
I almost understand nothing... It is not really my domain, but I thank you for the time you spend to try to explain me. You should have did it faster by yourself :-(
I have commented on #227

Ocramius:
> I almost understand nothing...

Just a little bit of patience :-)

 > It is not really my domain

You *really* need to make this your domain - testing is strictly needed for every software developer, so use this as a chance to get better at it :-)

Laurent3170:
I agree, you're right ! Testing is really needed and I do some (should do more, specifically for calling forms...) for web page calls, not for "applications". I have to progress, but there are not a lot of tutorials about creating tests (or I haven't found them, as I haven't really searched...)
I started with Symfony, by myself with tutotials, 18 months ago, and I use Doctrine only on a basic level. Working with cache is almost "chinese" for me, for the time being, even if cache is really important.

But again I will not be able to look at it before 3-4 weeks... Impossible to work with three young children plus 2 older ones ;-)

My main problem  with that test is that I don't know how to run the test and then being able to debug it. For Symfony apps I use `phpunit` but here, I don't know...

Thanks for your encouragements!

Ocramius:
@Laurent3170 this is also `phpunit`. Nobody is pushing you, just don't abandon it if you don't know how to continue: give it another shot when and if you have time.

It's OSS, there is nobody whipping you if you can't work on it immediately :-P

PR #229, opened 2017-08-16
add support for ext-couchbase:^2.3.0

### LINKED ISSUE #216: Couchbase support
hi Doctrine,

i'm looking for a PSR Couchbase implementation and came across a handful. of particular note is http://www.php-cache.com/en/latest/ which references Doctrine as being a Couchbase cache provider. i notice that the build does not install the extension, so tests are skipped. it appears that the adapter in this package expects an older version of the extension as well. current extension version(s) has \CouchbaseCluster and \CouchbaseBucket, or \Couchbase\Cluster and \Couchbase\Bucket instead of \Couchbase.

was wondering if i am on the right track or if i'm way off base. if i'm on to something, i was wondering if there is any interest in asserting the extension version in the existing adapter, and adding a new adapter for the more recent Couchbase extension. if so, i'd be interested in contributing.

abacaphiliac:
fixes #216 

hi Team,

this request adds CouchbaseBucketCache which requires `ext-couchbase:^2.3.0`.

i've updated the Travis CI build to install the Couchbase extension, and am using a Couchbase Server Docker image forked by WebPT (my employer) which auto-creates the default bucket with flush enabled.

ext-couchbase version is checked in the CouchbaseBucketCache constructor, by ensuring that the `manager` method is callable.

`ext-couchbase:^2.0,<2.3.0` does not yet have `$bucket->manager()` which is required to flush the cache pool and to pull cluster stats. we might be able to support older versions of Couchbase 2.x, but i think we might lose some functionality that is required by this library's API. 

Couchbase throws an exception on delete if a key is not found. That is handled gracefully in this new provider.

Couchbase Manager's flush operation returns void, so you can't tell if it was successful, e.g. it will fail silently if flush is not enabled for the bucket on the server. i set a marker key into cache so that the marker can be queried for success of the flush operation.

i updated the old provider to indicate it is deprecated, with a reference to the new provider and a link to the deprecation documentation from Couchbase.

i added a new test case for this new provider, and tightened the phpunit test case requirement to control which test case is executed based on which specific version of the extension is installed.

### DIFF
--- a/lib/Doctrine/Common/Cache/CouchbaseBucketCache.php
+++ b/lib/Doctrine/Common/Cache/CouchbaseBucketCache.php
@@ -30,13 +30,13 @@ use Couchbase\Exception;
  */
 final class CouchbaseBucketCache extends CacheProvider
 {
-    private CONST MINIMUM_VERSION = '2.3.0';
+    private const MINIMUM_VERSION = '2.3.0';
 
-    private CONST KEY_NOT_FOUND = 13;
+    private const KEY_NOT_FOUND = 13;
 
-    private CONST MAX_KEY_LENGTH = 250;
+    private const MAX_KEY_LENGTH = 250;
 
-    private CONST THIRTY_DAYS_IN_SECONDS = 2592000;
+    private const THIRTY_DAYS_IN_SECONDS = 2592000;
 
     /**
      * @var Bucket
@@ -91,7 +91,7 @@ final class CouchbaseBucketCache extends CacheProvider
         }
 
         if ($document instanceof Document) {
-            return !$document->error;
+            return ! $document->error;
         }
 
         return false;
@@ -117,7 +117,7 @@ final class CouchbaseBucketCache extends CacheProvider
         }
 
         if ($document instanceof Document) {
-            return !$document->error;
+            return ! $document->error;
         }
 
         return false;
@@ -137,7 +137,7 @@ final class CouchbaseBucketCache extends CacheProvider
         }
 
         if ($document instanceof Document) {
-            return !$document->error;
+            return ! $document->error;
         }
 
         return false;
@@ -170,10 +170,10 @@ final class CouchbaseBucketCache extends CacheProvider
      */
     protected function doGetStats()
     {
-        $manager = $this->bucket->manager();
-        $stats   = $manager->info();
-        $nodes = $stats['nodes'];
-        $node = $nodes[0];
+        $manager          = $this->bucket->manager();
+        $stats            = $manager->info();
+        $nodes            = $stats['nodes'];
+        $node             = $nodes[0];
         $interestingStats = $node['interestingStats'];
 
         return [
--- a/lib/Doctrine/Common/Cache/CouchbaseBucketCache.php
+++ b/lib/Doctrine/Common/Cache/CouchbaseBucketCache.php
@@ -17,6 +17,8 @@
  * <http://www.doctrine-project.org>.
  */
 
+declare(strict_types=1);
+
 namespace Doctrine\Common\Cache;
 
 use Couchbase\Bucket;
@@ -28,6 +30,8 @@ use Couchbase\Exception;
  */
 final class CouchbaseBucketCache extends CacheProvider
 {
+    private CONST MINIMUM_VERSION = '2.3.0';
+
     private CONST KEY_NOT_FOUND = 13;
 
     private CONST MAX_KEY_LENGTH = 250;
@@ -45,12 +49,12 @@ final class CouchbaseBucketCache extends CacheProvider
      */
     public function __construct(Bucket $bucket)
     {
-        $this->bucket = $bucket;
-
-        if (!is_callable([$bucket, 'manager'])) {
+        if (version_compare(phpversion('couchbase'), self::MINIMUM_VERSION) < 0) {
             // Manager is required to flush cache and pull stats.
-            throw new \RuntimeException('ext-couchbase:^2.3.0 is required.');
+            throw new \RuntimeException(sprintf('ext-couchbase:^%s is required.', self::MINIMUM_VERSION));
         }
+
+        $this->bucket = $bucket;
     }
 
     /**
@@ -67,7 +71,7 @@ final class CouchbaseBucketCache extends CacheProvider
         }
 
         if ($document instanceof Document && $document->value !== false) {
-            return $this->decode($document->value);
+            return unserialize($document->value);
         }
 
         return false;
@@ -103,7 +107,7 @@ final class CouchbaseBucketCache extends CacheProvider
         $lifeTime = $this->normalizeExpiry($lifeTime);
 
         try {
-            $encoded = $this->encode($data);
+            $encoded = serialize($data);
 
             $document = $this->bucket->upsert($id, $encoded, [
                 'expiry' => (int) $lifeTime,
@@ -197,31 +201,13 @@ final class CouchbaseBucketCache extends CacheProvider
     }
 
     /**
-     * @param mixed $value
-     * @return string
-     */
-    private function encode($value) : string
-    {
-        return serialize($value);
-    }
-
-    /**
-     * @param string $value
-     * @return mixed
-     */
-    private function decode(string $value)
-    {
-        return unserialize($value);
-    }
-
-    /**
      * Expiry treated as a unix timestamp instead of an offset if expiry is greater than 30 days.
      * @src https://developer.couchbase.com/documentation/server/4.1/developer-guide/expiry.html
      *
      * @param int $expiry
      * @return int
      */
-    private function normalizeExpiry($expiry) : int
+    private function normalizeExpiry(int $expiry) : int
     {
         if ($expiry > self::THIRTY_DAYS_IN_SECONDS) {
             return time() + $expiry;
--- a/lib/Doctrine/Common/Cache/CouchbaseBucketCache.php
+++ b/lib/Doctrine/Common/Cache/CouchbaseBucketCache.php
@@ -32,6 +32,8 @@ final class CouchbaseBucketCache extends CacheProvider
 
     private CONST MAX_KEY_LENGTH = 250;
 
+    private CONST THIRTY_DAYS_IN_SECONDS = 2592000;
+
     /**
      * @var Bucket
      */
@@ -98,9 +100,7 @@ final class CouchbaseBucketCache extends CacheProvider
     {
         $id = $this->normalizeKey($id);
 
-        if ($lifeTime > 30 * 24 * 3600) {
-            $lifeTime = time() + $lifeTime;
-        }
+        $lifeTime = $this->normalizeExpiry($lifeTime);
 
         try {
             $encoded = $this->encode($data);
@@ -185,7 +185,7 @@ final class CouchbaseBucketCache extends CacheProvider
      * @param string $id
      * @return string
      */
-    private function normalizeKey(string $id)
+    private function normalizeKey(string $id) : string
     {
         $normalized = substr($id, 0, self::MAX_KEY_LENGTH);
 
@@ -200,7 +200,7 @@ final class CouchbaseBucketCache extends CacheProvider
      * @param mixed $value
      * @return string
      */
-    private function encode($value) :string
+    private function encode($value) : string
     {
         return serialize($value);
     }
@@ -212,5 +212,21 @@ final class CouchbaseBucketCache extends CacheProvider
     private function decode(string $value)
     {
         return unserialize($value);
+    }
+
+    /**
+     * Expiry treated as a unix timestamp instead of an offset if expiry is greater than 30 days.
+     * @src https://developer.couchbase.com/documentation/server/4.1/developer-guide/expiry.html
+     *
+     * @param int $expiry
+     * @return int
+     */
+    private function normalizeExpiry($expiry) : int
+    {
+        if ($expiry > self::THIRTY_DAYS_IN_SECONDS) {
+            return time() + $expiry;
+        }
+
+        return $expiry;
     }
 }
--- a/lib/Doctrine/Common/Cache/CouchbaseBucketCache.php
+++ b/lib/Doctrine/Common/Cache/CouchbaseBucketCache.php
@@ -26,8 +26,12 @@ use Couchbase\Exception;
 /**
  * Couchbase ^2.3.0 cache provider.
  */
-class CouchbaseBucketCache extends CacheProvider
+final class CouchbaseBucketCache extends CacheProvider
 {
+    private CONST KEY_NOT_FOUND = 13;
+
+    private CONST MAX_KEY_LENGTH = 250;
+
     /**
      * @var Bucket
      */
@@ -57,7 +61,6 @@ class CouchbaseBucketCache extends CacheProvider
         try {
             $document = $this->bucket->get($id);
         } catch (Exception $e) {
-            // TODO Log exception.
             return false;
         }
 
@@ -78,7 +81,6 @@ class CouchbaseBucketCache extends CacheProvider
         try {
             $document = $this->bucket->get($id);
         } catch (Exception $e) {
-            // TODO Log exception.
             return false;
         }
 
@@ -107,7 +109,6 @@ class CouchbaseBucketCache extends CacheProvider
                 'expiry' => (int) $lifeTime,
             ]);
         } catch (Exception $e) {
-            // TODO Log exception.
             return false;
         }
 
@@ -128,12 +129,7 @@ class CouchbaseBucketCache extends CacheProvider
         try {
             $document = $this->bucket->remove($id);
         } catch (Exception $e) {
-            if ($e->getCode() === 13) {
-                return true;
-            }
-
-            // TODO Log exception.
-            return false;
+            return $e->getCode() === self::KEY_NOT_FOUND;
         }
 
         if ($document instanceof Document) {
@@ -173,7 +169,7 @@ class CouchbaseBucketCache extends CacheProvider
         $manager = $this->bucket->manager();
         $stats   = $manager->info();
         $nodes = $stats['nodes'];
-        $node = $nodes[0]; // TODO Support more than 1 node.
+        $node = $nodes[0];
         $interestingStats = $node['interestingStats'];
 
         return [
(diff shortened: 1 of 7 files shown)


lib/Doctrine/Common/Cache/CouchbaseBucketCache.php
```diff
@@ -0,0 +1,241 @@
...
+        }
+
+        try {
+            $encoded = $this->encode($data);
+
+            $document = $this->bucket->upsert($id, $encoded, [
+                'expiry' => (int) $lifeTime,
+            ]);
+        } catch (Exception $e) {
+            // TODO Log exception.
```
Ocramius: No logging needed

lib/Doctrine/Common/Cache/CouchbaseBucketCache.php
```diff
@@ -0,0 +1,241 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function doSave($id, $data, $lifeTime = 0)
+    {
+        $id = $this->normalizeKey($id);
+
+        if ($lifeTime > 30 * 24 * 3600) {
```
Ocramius: What is going on here?

lib/Doctrine/Common/Cache/CouchbaseBucketCache.php
```diff
@@ -0,0 +1,241 @@
...
+            $lifeTime = time() + $lifeTime;
```
Ocramius: Please don't overwrite variables if their semantic meaning changes

lib/Doctrine/Common/Cache/CouchbaseBucketCache.php
```diff
@@ -0,0 +1,241 @@
...
+        $id = $this->normalizeKey($id);
+
+        try {
+            $document = $this->bucket->remove($id);
+        } catch (Exception $e) {
+            if ($e->getCode() === 13) {
+                return true;
+            }
+
+            // TODO Log exception.
```
Ocramius: No need

lib/Doctrine/Common/Cache/CouchbaseBucketCache.php
```diff
@@ -0,0 +1,241 @@
...
+     * {@inheritdoc}
+     */
+    protected function doDelete($id)
+    {
+        $id = $this->normalizeKey($id);
+
+        try {
+            $document = $this->bucket->remove($id);
+        } catch (Exception $e) {
+            if ($e->getCode() === 13) {
```
Ocramius: `return $e->getCode() === 13`

Ocramius: Also, move `13` to a private constant, and give it a name

lib/Doctrine/Common/Cache/CouchbaseBucketCache.php
```diff
@@ -0,0 +1,241 @@
...
+namespace Doctrine\Common\Cache;
+
+use Couchbase\Bucket;
+use Couchbase\Document;
+use Couchbase\Exception;
+
+/**
+ * Couchbase ^2.3.0 cache provider.
+ */
+class CouchbaseBucketCache extends CacheProvider
```
Ocramius: `final`

lib/Doctrine/Common/Cache/CouchbaseBucketCache.php
```diff
@@ -0,0 +1,241 @@
...
+    {
+        $manager = $this->bucket->manager();
+
+        // Flush does not return with success or failure, and must be enabled per bucket on the server.
+        // Store a marker item so that we will know if it was successful.
+        $this->doSave(__METHOD__, true, 60);
+
+        $manager->flush();
+
+        if ($this->doContains(__METHOD__)) {
```
Ocramius: Using `__METHOD__` is only safe if the implementation is `final`

lib/Doctrine/Common/Cache/CouchbaseBucketCache.php
```diff
@@ -0,0 +1,241 @@
...
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function doGetStats()
+    {
+        $manager = $this->bucket->manager();
+        $stats   = $manager->info();
+        $nodes = $stats['nodes'];
+        $node = $nodes[0]; // TODO Support more than 1 node.
```
Ocramius: Is this worth doing here? Usually, a reverse HTTP proxy does this stuff, no?

lib/Doctrine/Common/Cache/CouchbaseBucketCache.php
```diff
@@ -0,0 +1,241 @@
...
+            Cache::STATS_MEMORY_USAGE     => $interestingStats['mem_used'],
+            Cache::STATS_MEMORY_AVAILABLE => $node['memoryFree'],
+        ];
+    }
+
+    /**
+     * @param $id
+     * @return bool|string
+     */
+    private function normalizeKey($id)
```
Ocramius: For all private API, please add parameter and return type hints (PHP 7.1 style is OK)

lib/Doctrine/Common/Cache/CouchbaseBucketCache.php
```diff
@@ -0,0 +1,241 @@
...
+    {
+        $normalized = substr($id, 0, 250);
+
+        if ($normalized === false) {
+            // TODO Log it?
```
Ocramius: No - test it :-)

lib/Doctrine/Common/Cache/CouchbaseBucketCache.php
```diff
@@ -0,0 +1,241 @@
...
+        return $normalized;
+    }
+
+    /**
+     * @param mixed $value
+     * @return mixed
+     */
+    private function encode($value)
+    {
+        if (is_array($value) || is_object($value) || is_numeric($value)) {
```
Ocramius: Consider ordering these checks by frequency

lib/Doctrine/Common/Cache/CouchbaseBucketCache.php
```diff
@@ -0,0 +1,241 @@
...
+        return json_encode($value);
+    }
+
+    /**
+     * @param mixed $value
+     * @return mixed
+     */
+    private function decode($value)
+    {
+        $decoded = @unserialize($value);
```
Ocramius: I wouldn't silence a deserialization operation here

lib/Doctrine/Common/Cache/CouchbaseBucketCache.php
```diff
@@ -0,0 +1,241 @@
...
+
+        if ($decoded !== false) {
+            return $decoded;
+        }
+
+        $decoded = json_decode($value);
```
Ocramius: Wait - please try being consistent about serialization formats: either it's json, or it's serialized. Not both, please.

tests/Doctrine/Tests/Common/Cache/CouchbaseBucketCacheTest.php
```diff
@@ -0,0 +1,32 @@
...
+{
+    /** @var Bucket */
+    private $bucket;
+
+    protected function setUp() : void
+    {
+        try {
+            $cluster = new Cluster('couchbase://localhost?detailed_errcodes=1');
+            $this->bucket = $cluster->openBucket('default');
+        } catch(\Exception $ex) {
```
Ocramius: Let it bubble up

abacaphiliac:
thank you! i will fix these tonight.

lib/Doctrine/Common/Cache/CouchbaseBucketCache.php
abacaphiliac: originally, i just copied from [\Doctrine\Common\Cache\CouchbaseCache](https://github.com/doctrine/cache/blob/fe68e4f78804684e31cce17590b0bd9f088db1b0/lib/Doctrine/Common/Cache/CouchbaseCache.php#L81), but then i saw in the PHPStorm `\Couchbase\Bucket::upsert()` stub that expiry > 30 days is treated in a special way.

here are the real [Couchbase expiry docs](https://developer.couchbase.com/documentation/server/4.1/developer-guide/expiry.html), which say:
```
If the absolute value of the expiry is less than 30 days (i.e. 60 * 60 * 24 * 30), it is considered an offset. If it is greater, it is considered an absolute time stamp.
```

Couchbase Server uses Memcached internally, and exposes a Memcached API to connect to specially configured buckets via their Moxi service. i think that the strange expiry behavior in Couchbase is meant to maintain compatibility with the Memcached API.

you can see the same special handling of expiry in the PHP Memcached docs and in the Memcached cache providers in this library:
* http://php.net/manual/en/memcache.set.php
* http://php.net/manual/en/memcached.expiration.php
* https://github.com/doctrine/cache/blob/206b6346465133b5939ffcf255d26559679a1da7/lib/Doctrine/Common/Cache/MemcachedCache.php#L107
* https://github.com/doctrine/cache/blob/fe68e4f78804684e31cce17590b0bd9f088db1b0/lib/Doctrine/Common/Cache/MemcacheCache.php#L89

lib/Doctrine/Common/Cache/CouchbaseBucketCache.php
abacaphiliac: per the docs, if `expiry` is greater than 30 days then it is treated as a unix timestamp instead of an offset : /

lib/Doctrine/Common/Cache/CouchbaseBucketCache.php
abacaphiliac: class is now final : )

lib/Doctrine/Common/Cache/CouchbaseBucketCache.php
abacaphiliac: could be. should we expect that everyone will have this setup? should we document the limitation? we only get accurate stats then if they use a reverse proxy. if we don't assume one is used, could we agree upon some aggregation logic here, e.g. simply adding the values? maybe it's not worth it. you tell me : )

lib/Doctrine/Common/Cache/CouchbaseBucketCache.php
abacaphiliac: done, except for `mixed` in the serialization functions. any suggestions for handling that pseudo-type?

lib/Doctrine/Common/Cache/CouchbaseBucketCache.php
abacaphiliac: i am struggling to test this, as type-checking for string happens in several private methods (including in the parent class). can it be considered dead code?

lib/Doctrine/Common/Cache/CouchbaseBucketCache.php
abacaphiliac: how embarrassing. thank you for calling this out. i had some other bugs in the class earlier that evening which were resolved by mixing formats. after those were resolved this blend was no longer necessary.

lib/Doctrine/Common/Cache/CouchbaseBucketCache.php
lcobucci: We can also extract `30 * 24 * 3600` to a private constant to make it more understandable as well

lib/Doctrine/Common/Cache/CouchbaseBucketCache.php
```diff
@@ -186,15 +182,14 @@ protected function doGetStats()
     }
 
     /**
-     * @param $id
-     * @return bool|string
+     * @param string $id
+     * @return string
      */
-    private function normalizeKey($id)
+    private function normalizeKey(string $id)
```
lcobucci: It should be `private function normalizeKey(string $id) : string` instead, right?

lib/Doctrine/Common/Cache/CouchbaseBucketCache.php
```diff
@@ -203,39 +198,19 @@ private function normalizeKey($id)
 
     /**
      * @param mixed $value
-     * @return mixed
+     * @return string
      */
-    private function encode($value)
+    private function encode($value) :string
```
lcobucci: We're using a whitespace before and after the colon, can you please change it?

lib/Doctrine/Common/Cache/CouchbaseBucketCache.php
abacaphiliac: great suggestions. sorry for misunderstanding.

lib/Doctrine/Common/Cache/CouchbaseBucketCache.php
abacaphiliac: ah of course!

abacaphiliac:
addressed most recent feedback, but now of course installation of the couchbase extension is failing.

abacaphiliac:
rebuild passed:
https://travis-ci.org/abacaphiliac/cache/builds/265616518

i'm nervous to make the build brittle. i've reached out to Travis CI about adding the Couchbase extension and service.

lib/Doctrine/Common/Cache/CouchbaseBucketCache.php
```diff
@@ -0,0 +1,232 @@
...
+
+    /**
+     * CouchbaseCache constructor.
+     * @param Bucket $bucket
+     */
+    public function __construct(Bucket $bucket)
+    {
+        $this->bucket = $bucket;
+
+        if (!is_callable([$bucket, 'manager'])) {
```
lcobucci: Can you please validate this BEFORE of assigning the property? And wouldn't it be better (more reliable) to use something like `version_compare(phpversion('couchbase'), '2.3.0') >= 0`?

lcobucci: Also, `if ( ! is_callable([$bucket, 'manager'])) {`

lib/Doctrine/Common/Cache/CouchbaseBucketCache.php
```diff
@@ -0,0 +1,232 @@
...
+    }
+
+    /**
+     * Expiry treated as a unix timestamp instead of an offset if expiry is greater than 30 days.
+     * @src https://developer.couchbase.com/documentation/server/4.1/developer-guide/expiry.html
+     *
+     * @param int $expiry
+     * @return int
+     */
+    private function normalizeExpiry($expiry) : int
```
lcobucci: `int $expiry`

lib/Doctrine/Common/Cache/CouchbaseBucketCache.php
```diff
@@ -0,0 +1,232 @@
...
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ * This software consists of voluntary contributions made by many individuals
+ * and is licensed under the MIT license. For more information, see
+ * <http://www.doctrine-project.org>.
+ */
+
```
lcobucci: Can you please add `declare(strict_types=1);` here?

lib/Doctrine/Common/Cache/CouchbaseBucketCache.php
```diff
@@ -0,0 +1,232 @@
...
+        }
+
+        return $normalized;
+    }
+
+    /**
+     * @param mixed $value
+     * @return string
+     */
+    private function encode($value) : string
```
lcobucci: I feel like this is not really needed anymore

lib/Doctrine/Common/Cache/CouchbaseBucketCache.php
```diff
@@ -0,0 +1,232 @@
...
+    {
+        return serialize($value);
+    }
+
+    /**
+     * @param string $value
+     * @return mixed
+     */
+    private function decode(string $value)
```
lcobucci: This one as well

tests/Doctrine/Tests/Common/Cache/CouchbaseCacheTest.php
```diff
@@ -7,7 +7,8 @@
 use Doctrine\Common\Cache\CouchbaseCache;
 
 /**
- * @requires extension couchbase
+ * @requires extension couchbase >=1.0
```
lcobucci: Will this test run on Travis-CI with the new config (I'm not even sure if it was running before 😂)?

lib/Doctrine/Common/Cache/CouchbaseBucketCache.php
abacaphiliac: done.

lcobucci:
@abacaphiliac I think it's looking good, can you rebase your branch with `master` to sync the latest changes and solve conflicts? You can also use `./vendor/bin/phpcbf` to fix potential coding standard violations 😉

tests/Doctrine/Tests/Common/Cache/CouchbaseCacheTest.php
abacaphiliac: you're correct; these tests did not run before, because no version of the extension was installed.

it seemed like a good idea to be very specific in this old test case since the new test case is also very specific.

we might want to run a different build with an older version of the extension, some day. not sure. do you think i should revert this annotation?

lcobucci: I'm not sure if this adapter is still relevant.

According to https://pecl.php.net/package-changelog.php?package=couchbase PHP 7 support has only been added on `2.2.0` and `doctrine/cache` now requires PHP 7.1 (so that excludes couchbase `v1`).

I've added https://github.com/doctrine/cache/issues/234 to discuss things like this.

abacaphiliac: excellent!

abacaphiliac:
@lcobucci rebased.

lcobucci:
@abacaphiliac great job, thanks for your contribution 😉

Ocramius:
:ship:

abacaphiliac:
was my pleasure. thank you @Ocramius and @lcobucci for taking the time to work through this with me. also, great job on the CI enhancements. very helpful : )

PR #230, opened 2017-08-24
Add PHPCS to the project

lcobucci:
To make sure everything follows our coding standards.


lcobucci:
@TomasVotruba here you go 😉

.travis.yml
```diff
@@ -20,19 +21,29 @@ before_install:
...
+    - stage: Coding standard
+      script:
+        - ./vendor/bin/phpcs
+
+    - stage: Coverage
+      before_script:
+        - mv ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/xdebug.ini{.disabled,}
+        - if [[ ! $(php -m | grep -si xdebug) ]]; then echo "xdebug required for coverage"; exit 1; fi
+      script:
+        - ./vendor/bin/phpunit --exclude-group performance --coverage-clover clover.xml
```
Ocramius: Since you rely on phpcs, phpcs should also be a dependency (besides `doctrine/coding-standard`)

lcobucci: Alright, will add it then.

tests/Doctrine/Tests/Common/Cache/CacheTest.php
```diff
@@ -128,9 +128,9 @@ public function testSaveMultiple() : void
 
     public function provideDataToCache() : array
     {
-        $obj = new \stdClass();
-        $obj->foo = 'bar';
-        $obj2 = new \stdClass();
+        $obj       = new \stdClass();
```
TomasVotruba: I wonder, this is part of Doctrine cs? Since when?

phpcs.xml.dist
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0"?>
```
TomasVotruba: This file might be used from Doctrine/CodingStandard, right? This might be unnecessary duplication

TomasVotruba: Then loading it in CLI with `--config`

tests/Doctrine/Tests/Common/Cache/CacheTest.php
TomasVotruba: This might lead to reformating many line just due to adding a new variable.

Ocramius: Yes, absolutely, but we value readability over conflicts, I'd say.

We always required to align these. If I didn't ask it to the committers, I did it myself when merging patches.

Ocramius: I don't mind the additional diff as long as the code block is actually understandable later on.

TomasVotruba: If that would be more readable, I'd be ok with git conflicts. It lead to read in columns, not in lines, which is counter intuitive.

To be honest, I haven't seen this type of code for a while. I see it in Doctrine first time, that's why I'm surprised.

Ocramius: Yeah, I use it with all my libs regardless, because from a logical perspective, we're assigning a bunch of stuff all together

phpcs.xml.dist
lcobucci: I'd say we should have a `phpcs.xml.dist` for each project, the same way each project has it's own `phpunit.xml.dist`.

The file on `doctrine/coding-standard` is to validate `doctrine/coding-standard` code.

Majkl578: Agreed, each project should have its own dist file.

PR #233, opened 2017-08-25
Improve SQLite3 adapter

lcobucci:
Tests were simply not being executed because PHPUnit was considering `sqlite` as extension name (instead of `sqlite3`) and we also had a weird `return` in the constructor.


Ocramius:
:ship:

PR #235, opened 2017-08-25
Upgrade PHPUnit

lcobucci:
Squeezing a small fix in 😆

### DIFF
--- a/tests/Doctrine/Tests/Common/Cache/ArrayCacheTest.php
+++ b/tests/Doctrine/Tests/Common/Cache/ArrayCacheTest.php
@@ -28,11 +28,11 @@ class ArrayCacheTest extends CacheTest
         $cache->fetch('test3');
 
         $stats = $cache->getStats();
-        $this->assertEquals(2, $stats[Cache::STATS_HITS]);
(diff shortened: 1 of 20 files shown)


Ocramius:
:ship:

tests/Doctrine/Tests/Common/Cache/ArrayCacheTest.php
```diff
@@ -28,11 +28,11 @@ public function testGetStats() : void
         $cache->fetch('test3');
 
         $stats = $cache->getStats();
-        $this->assertEquals(2, $stats[Cache::STATS_HITS]);
-        $this->assertEquals(5, $stats[Cache::STATS_MISSES]); // +1 for internal call to DoctrineNamespaceCacheKey
-        $this->assertNotNull($stats[Cache::STATS_UPTIME]);
-        $this->assertNull($stats[Cache::STATS_MEMORY_USAGE]);
-        $this->assertNull($stats[Cache::STATS_MEMORY_AVAILABLE]);
+        self::assertEquals(2, $stats[Cache::STATS_HITS]);
```
Tobion: Why? It has always been static but that is not how the assertions are used commonly.

Ocramius: We already switched most of the test suites to use the correct scoping rules. This is simply more correct.

PR #237, opened 2017-10-16
Bloom Filter Suggestion

igoreus:
@Ocramius 
There is an idea to add in front of chainCache Bloom filter. It's not final implementation. I just would like to know your opinion and does it make sense to continue.

Why it might be useful:
- If chainCache has long list of caches.
- To reduce amount of requests to backend
- If bloom-filter will be used in front of slow storage
Probably will not help if  chain is short and fast e.g. array + redis. 

If the idea make sense then "counting bloom filter" or "cuckoo filter" should be implemented to support delete functionality.


lib/Doctrine/Common/Cache/BloomChainCache.php
```diff
@@ -0,0 +1,126 @@
...
+
+use RocketLabs\BloomFilter\BloomFilterAbstract;
+
+
+/**
+ * Cache provider that uses bloom filter in front of other cache providers
+ *
+ * @author Igor Veremchuk <igor.veremchuk@gmail.com>
+ */
+class BloomChainCache extends ChainCache
```
Ocramius: `final`

Ocramius: We should probably design this without the chain cache in mind

lib/Doctrine/Common/Cache/BloomChainCache.php
```diff
@@ -0,0 +1,126 @@
...
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ * This software consists of voluntary contributions made by many individuals
+ * and is licensed under the MIT license. For more information, see
+ * <http://www.doctrine-project.org>.
+ */
+
+namespace Doctrine\Common\Cache;
+
+use RocketLabs\BloomFilter\BloomFilterAbstract;
```
Ocramius: How complex is the bloom filter implementation? Could it be ported to avoid a direct dependency?

lib/Doctrine/Common/Cache/BloomChainCache.php
```diff
@@ -0,0 +1,126 @@
...
+class BloomChainCache extends ChainCache
+{
+    /** @var BloomFilterAbstract */
+    protected $bloomFilter;
+
+    /**
+     * @param BloomFilterAbstract $bloomFilter
+     * @param array $cacheProviders
+     */
+    public function __construct(BloomFilterAbstract $bloomFilter , $cacheProviders = [])
```
Ocramius: ` __construct(BloomFilterAbstract $bloomFilter , Cache $firstCacheDriver, Cache ...$otherCacheDrivers)`

lib/Doctrine/Common/Cache/NotSupportedException.php
```diff
@@ -0,0 +1,26 @@
...
+ * and is licensed under the MIT license. For more information, see
+ * <http://www.doctrine-project.org>.
+ */
+
+namespace Doctrine\Common\Cache;
+
+/**
+ * @author Igor Veremchuk igor.veremchuk@gmail.com
+ */
+class NotSupportedException extends \BadMethodCallException
```
Ocramius: Probably needs better naming

Ocramius (changes requested):
This is a brilliant approach for very high miss caches with slow backends. Also, implementation is quite simple :+1: 

Overall, the implementation requires discussion and a lot of honing (I would start from just the interface implementation), but this seems to be quite useful for a lot of people.

lib/Doctrine/Common/Cache/BloomChainCache.php
igoreus: It's not so complex. I will  port it and we will see

lib/Doctrine/Common/Cache/BloomChainCache.php
igoreus: I also think it would be better. 
But should it be in "doctrine/cache" or "doctrine/bloom"  
How would you prefer to use it? maybe you have some use cases right now

lib/Doctrine/Common/Cache/NotSupportedException.php
igoreus: I think to remove it at all but I need it for now until I implement deleting functionality

igoreus:
@Ocramius 
Please have a look

There is 3 implementations of Bloom filter:
-

- Bloom Filter
    - Pros:
        - Fastest filter
        - Use less memory than other implementations
    - Cons:
        - Does not support deletion
        - Does not support resize. Max amount of set members should be known in advance.
- Dynamic Bloom Filter
     - Pros:
        - Amount of members in dynamic
     - Cons:
         - Does not support deletion
         - Slower than Bloom Filter
         - False positive probability increase each time when the filter need to be resized

- Counting Bloom Filter
    - Pros:
        - Supports deleting from filter
    - Cons:
        - Needs more memory. Additional 4 bits or 8 bits per flag
        - Does not support resize. Max amount of set members should be known in advance.

Supported Bloom Filter storages:
- 
 - In Memory 
 - in Redis
 
Supported Hashes
-
 - Murmur3: Better uniform distribution.
 - Crc32b: Faster computation.

Cache Chain - Example of usage
-

```php
use Doctrine\Common\Cache\RedisCache;
use Doctrine\Common\Cache\ArrayCache;
use Doctrine\Common\Cache\ChainCache;
use Doctrine\Common\Cache\BloomChainCache;
use Doctrine\Common\Filter\Persist\BitRedis;
use Doctrine\Common\Filter\DynamicBloomFilter;
use Doctrine\Common\Filter\Hash\Murmur;


$redis = new Redis();
$redis->connect('localhost');

// init caches
$redisCache = new RedisCache();
$redisCache->setRedis($redis);
$arrayCache = new ArrayCache();
$chainCache = new ChainCache([$arrayCache, $redisCache]);

// init Bloom Filter
$persisterRedis = new BitRedis($redis, 'dynamic_bloom_filter');
$hashFunction = new Murmur();
$bloomFilter = new DynamicBloomFilter($persisterRedis, $hashFunction);

//restore from memento if exists
//get memento from storage
if (isset($memento)) {
    $bloomFilter->restoreState($memento);
} else {
    $bloomFilter->setSize(10000);
}

//Create bloom chain cache
$bloomChainCache = new BloomChainCache($bloomFilter, $arrayCache, $redisCache);


$bloomChainCache->save('test1', 'data1');

$bloomChainCache->fetch('test1'); // exists in bloom, continue to chainCache
$bloomChainCache->fetch('test2'); // does not exist in bloom. return false


//Save State between requests
$memento = $bloomFilter->saveState();
//save $memento to some storage

```

Counting Bloom Filter Example of usage
-
```php

use Doctrine\Common\Filter\Persist\BitRedis;
use Doctrine\Common\Filter\CountingBloomFilter;
use Doctrine\Common\Filter\Hash\Murmur;


$redis = new Redis();
$redis->connect('localhost');

// init Bloom Filter
$persisterRedis = new BitRedis($redis, 'dynamic_bloom_filter');
$hashFunction = new Murmur();
$bloomFilter = new DynamicBloomFilter($persisterRedis, $hashFunction);

$bloomFilter->add('test1');
$bloomFilter->has('test1'); // returns true //Probably in filter
$bloomFilter->delete('test1');
$bloomFilter->has('test1'); // returns false //Defenetly not in filter
```

Tobion:
To me it seems like a perfect fit as a standalone package. I don't see why it should be part of doctrine/cache itself. Just release this as a separate package on top.

igoreus:
@Ocramius have you had a chance to have a look. Would be nice if we can discuss next steps to move forward?
- Most probably I should create standalone bloomfilter  package as @Tobion suggested.
- Can we align possible use cases to create proper interfaces?
- Should I ask other developers to review the code?

Ocramius:
Didn't have a chance to look at it further, sorry.

Marco Pivetta

http://twitter.com/Ocramius

http://ocramius.github.com/

alcaeus:

PR #238, opened 2017-11-20
Cache namespace version in own cache

mikemeier:
Actually every request generates one missing key (current namespaceCacheKey) as long as no "deleteAll" is called.

### DIFF
--- a/lib/Doctrine/Common/Cache/CacheProvider.php
+++ b/lib/Doctrine/Common/Cache/CacheProvider.php
@@ -216,9 +216,14 @@ abstract class CacheProvider implements Cache, FlushableCache, ClearableCache, M
         if (null !== $this->namespaceVersion) {
             return $this->namespaceVersion;
         }
+        
+        $namespaceCacheKey = $this->getNamespaceCacheKey();
+        if($this->doContains($namespaceCacheKey)){
+            return $this->namespaceVersion = (int) $this->doFetch($namespaceCacheKey);
+        }
 
-        $namespaceCacheKey      = $this->getNamespaceCacheKey();
-        $this->namespaceVersion = (int) $this->doFetch($namespaceCacheKey) ?: 1;
+        $this->namespaceVersion = 1;
+        $this->doSave($namespaceCacheKey, $this->namespaceVersion);
 
         return $this->namespaceVersion;
     }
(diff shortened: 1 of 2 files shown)


tests/Doctrine/Tests/Common/Cache/CacheProviderTest.php
```diff
@@ -143,4 +143,22 @@ public function testInvalidNamespaceVersionCacheEntry() : void
...
+        $cache->expects($this->once())
+              ->method('doContains')
+              ->willReturn(false);
+
+        $cache->expects($this->exactly(2))
+              ->method('doFetch')
+              ->with('[missing_key][1]')
+              ->willReturn(null);
+
+        $cache->fetch('missing_key');
```
Ocramius: Are the return types here always `null`?

tests/Doctrine/Tests/Common/Cache/CacheProviderTest.php
```diff
@@ -143,4 +143,23 @@ public function testInvalidNamespaceVersionCacheEntry() : void
 
         self::assertTrue($cache->deleteAll());
     }
+
+    public function testFetchWillNotResultInMissingCacheKeyAfterFirstTry(): void
```
Ocramius: Can you rename the method so it exposes what you are fixing? The method body seems to attempt two `doFetch` oprations and a `doContains`, but no namespace-related operations seem to be happening.

lib/Doctrine/Common/Cache/CacheProvider.php
```diff
@@ -216,9 +216,14 @@ private function getNamespaceVersion() : int
         if (null !== $this->namespaceVersion) {
             return $this->namespaceVersion;
         }
+        
+        $namespaceCacheKey = $this->getNamespaceCacheKey();
+        if($this->doContains($namespaceCacheKey)){
```
alcaeus: Nitpick: space after `if` and before `{`

Tobion:
This basically reverts #91. I'm fine with it since it has been reported before already in #141

alcaeus:
@mikemeier can you fix issues mentioned in reviews? In particular, there need to be more tests to cover your changes. Thanks!

mikemeier:
@alcaeus Looks like this "problem" is already discussed in https://github.com/doctrine/cache/issues/141

PR #253, opened 2018-03-19
Dont "use function time", it prevents mocking time()

nicolas-grekas:


Ocramius:
This looks wrong, IMO. If we'd have to mock time (which we don't in this package), we'd inject a clock (https://blog.frankdejonge.nl/being-in-control-of-time-in-php/).

Still, the cache component does not need to do that.

I suggest mocking the cache interface instead.

nicolas-grekas:
It's quite convenient, see https://symfony.com/blog/new-in-symfony-2-8-clock-mocking-and-time-sensitive-tests
But I'll do what you suggest.

Ocramius:
@nicolas-grekas even without the specific detail of `time()`, this is cutting dangerously deep into a third party component, so it seems unsafe overall.

Doing stuff like this will likely always turn into problems as we move internal components on our side.

Cool idea, but it really is touching things it shouldn't...

Majkl578:
I wouldn't call code injection into other vendor's namespace _mocking_. People find this convenient but it's just abuse of function fallback lookups and could be dangerous.
Actually this is one of the reasons why I encourage using fully qualified functions (either by backslash or `use function`).

There is also (runkit_function_redefine)[http://php.net/manual/en/function.runkit-function-redefine.php] if anyone _really_ needs that.

nicolas-grekas:
I understand you have this pov. We'll use a simple mock implementation in the component's test suite.

PR #259, opened 2018-05-17
Fix Race Condition in RedisCache::doFetchMultiple

### LINKED ISSUE #258: Race Condition in RedisCache::doFetchMultiple
RedisCache calls "mget" to fetch multiple values: https://github.com/doctrine/cache/blob/master/lib/Doctrine/Common/Cache/RedisCache.php#L60

If one of the returned values is false then it double-checks whether the value really doesn't exist by calling the "exists" method on the key. Now, if the value has just been (re-)created between the "mget" and the "exists" call then RedisCache thinks that false really is the value of the specified key (although it isn't, it just called "mget" while the key didn't exist, then later called "exists" after the key was re-added).

I think this is a conceptional issue of letting PhpRedis serialize the data (instead of doing that in the RedisCache library). With serialization enabled, PhpRedis returns false for a) keys that don't exist and b) for keys that exist and have the value false. To distinguish between these two cases, RedisCache uses the "exists" call which unfortunately introduces the described race condition.

I think RedisCache should do the (de-)serialization on its own.

This issue seems to have been introduced in https://github.com/doctrine/cache/commit/47f3e599160fb3fce74072c47649b7dbfdd1e72d.

Reference: https://xenforo.com/community/threads/argument-1-passed-to-xf-template-templater-setgroupstyles-must-be-of-the-type-array.139491/#post-1252143

Tobion:
Fixes #258

### DIFF
--- a/lib/Doctrine/Common/Cache/RedisCache.php
+++ b/lib/Doctrine/Common/Cache/RedisCache.php
@@ -57,8 +57,11 @@ class RedisCache extends CacheProvider
         $foundItems = [];
 
         foreach ($fetchedItems as $key => $value) {
-            if ($value === false && ! $this->redis->exists($key)) {
-                continue;
+            if ($value === false) {
+                if ( ! $this->redis->exists($key)) {
+                    continue;
+                }
+                $value = $this->redis->get($key);
             }
 
             $foundItems[$key] = $value;


lib/Doctrine/Common/Cache/RedisCache.php
```diff
@@ -57,8 +57,11 @@ protected function doFetchMultiple(array $keys)
         $foundItems = [];
 
         foreach ($fetchedItems as $key => $value) {
-            if ($value === false && ! $this->redis->exists($key)) {
-                continue;
+            if ($value === false) {
+                if ( ! $this->redis->exists($key)) {
```
Majkl578: There is still race condition between `->exists()` and `->get()` calls, this would require different approach (either MULTI or relying on GET only).

Tobion: Nvm, this just reduces the likeliness but does indeed not remove the race condition.

PR #260, opened 2018-05-17
Fix Race Condition in RedisCache::doFetchMultiple

### LINKED ISSUE #258: Race Condition in RedisCache::doFetchMultiple

steffenweber:
Fixes #258

### DIFF
--- a/lib/Doctrine/Common/Cache/RedisCache.php
+++ b/lib/Doctrine/Common/Cache/RedisCache.php
@@ -15,8 +15,14 @@ use function is_bool;
  */
 class RedisCache extends CacheProvider
 {
+    public const SERIALIZER_PHP = 1;
+    public const SERIALIZER_IGBINARY = 2;
+
     /** @var Redis|null */
     private $redis;
+
+    /** @var int|null */
+    private $serializer;
 
     /**
      * Sets the redis instance to use.
@@ -25,8 +31,8 @@ class RedisCache extends CacheProvider
      */
     public function setRedis(Redis $redis)
     {
-        $redis->setOption(Redis::OPT_SERIALIZER, $this->getSerializerValue());
         $this->redis = $redis;
+        $this->serializer = $this->getSerializerValue();
     }
 
     /**
@@ -44,7 +50,13 @@ class RedisCache extends CacheProvider
      */
     protected function doFetch($id)
     {
-        return $this->redis->get($id);
+        $value = $this->redis->get($id);
+
+        if (false === $value) {
+            return false;
+        }
+
+        return $this->unserialize($value);
     }
 
     /**
@@ -58,11 +70,11 @@ class RedisCache extends CacheProvider
         $foundItems = [];
 
         foreach ($fetchedItems as $key => $value) {
-            if ($value === false && ! $this->redis->exists($key)) {
+            if ($value === false) {
                 continue;
             }
 
-            $foundItems[$key] = $value;
+            $foundItems[$key] = $this->unserialize($value);
         }
 
         return $foundItems;
@@ -78,7 +90,7 @@ class RedisCache extends CacheProvider
 
             // Keys have lifetime, use SETEX for each of them
             foreach ($keysAndValues as $key => $value) {
-                if ($this->redis->setex($key, $lifetime, $value)) {
+                if ($this->redis->setex($key, $lifetime, $this->serialize($value))) {
                     continue;
                 }
 
@@ -89,7 +101,9 @@ class RedisCache extends CacheProvider
         }
 
         // No lifetime, use MSET
-        return (bool) $this->redis->mset($keysAndValues);
+        return (bool) $this->redis->mset(
+            array_map([$this, 'serialize'], $keysAndValues)
+        );
     }
 
     /**
@@ -112,10 +126,10 @@ class RedisCache extends CacheProvider
     protected function doSave($id, $data, $lifeTime = 0)
     {
         if ($lifeTime > 0) {
-            return $this->redis->setex($id, $lifeTime, $data);
+            return $this->redis->setex($id, $lifeTime, $this->serialize($data));
         }
 
-        return $this->redis->set($id, $data);
+        return $this->redis->set($id, $this->serialize($data));
     }
 
     /**
@@ -157,6 +171,24 @@ class RedisCache extends CacheProvider
         ];
     }
 
+    protected function serialize($value)
+    {
+        if ($this->serializer === self::SERIALIZER_IGBINARY) {
+            return igbinary_serialize($value);
+        }
+
+        return serialize($value);
+    }
+
+    protected function unserialize($value)
+    {
+        if ($this->serializer === self::SERIALIZER_IGBINARY) {
+            return igbinary_unserialize($value);
+        }
+
+        return unserialize($value);
+    }
+
     /**
      * Returns the serializer constant to use. If Redis is compiled with
      * igbinary support, that is used. Otherwise the default PHP serializer is
@@ -167,9 +199,9 @@ class RedisCache extends CacheProvider
     protected function getSerializerValue()
     {
         if (defined('Redis::SERIALIZER_IGBINARY') && extension_loaded('igbinary')) {
-            return Redis::SERIALIZER_IGBINARY;
+            return self::SERIALIZER_IGBINARY;
         }
 
-        return Redis::SERIALIZER_PHP;
+        return self::SERIALIZER_PHP;
     }
 }
--- a/tests/Doctrine/Tests/Common/Cache/RedisCacheTest.php
+++ b/tests/Doctrine/Tests/Common/Cache/RedisCacheTest.php
@@ -47,9 +47,10 @@ class RedisCacheTest extends CacheTest
             $this->markTestSkipped('Extension igbinary is loaded.');
         }
 
-        self::assertEquals(
-            Redis::SERIALIZER_PHP,
-            $this->_getCacheDriver()->getRedis()->getOption(Redis::OPT_SERIALIZER)
+        self::assertAttributeSame(
+            RedisCache::SERIALIZER_PHP,
+            'serializer',
+            $this->_getCacheDriver()
         );
     }
 


lib/Doctrine/Common/Cache/RedisCache.php
```diff
@@ -149,20 +155,4 @@ protected function doGetStats()
...
-     * Returns the serializer constant to use. If Redis is compiled with
-     * igbinary support, that is used. Otherwise the default PHP serializer is
-     * used.
-     *
-     * @return int One of the Redis::SERIALIZER_* constants
-     */
-    protected function getSerializerValue()
-    {
-        if (defined('Redis::SERIALIZER_IGBINARY') && extension_loaded('igbinary')) {
-            return Redis::SERIALIZER_IGBINARY;
```
Ocramius, line 170: The patch breaks BC for those using `Redis::SERIALIZER_IGBINARY`

Ocramius (changes requested):

lib/Doctrine/Common/Cache/RedisCache.php
steffenweber: I think I may have found a solution. Could you have a look?

(I can later on squash all commits into a single commit.)

steffenweber: Afaik igbinary is a drop-in replacement for PHP's serialize/unserialize functions. So if you have igbinary installed then serialize/unserialize will automatically use the igbinary format. Hence, the new RedisCache will automatically use igbinary if it's available (just like the old RedisCache). So as far as I understand it there shouldn't be a BC break.

Xon: That only occurs when using php native serialization infrastructure not the serialize/unserialize functions.

steffenweber: I've added a commit that should fix this.

steffenweber:
Some tests were failing for reasons that seemed unrelated to my changes. I've therefore rebased my patch on the master branch but now tests are failing with the error "The command "./tests/travis/Couchbase.sh" failed and exited with 1 during."

Can someone let me know how we can continue? I think the patch is fine but it obviously isn't a one-liner and probably needs a review.

Majkl578:
> ut now tests are failing with the error "The command "./tests/travis/Couchbase.sh" failed and exited with 1 during"

See #271 for that.

lib/Doctrine/Common/Cache/RedisCache.php
```diff
@@ -157,6 +180,24 @@ protected function doGetStats()
         ];
     }
 
+    protected function serialize($value)
+    {
+        if ($this->serializer === self::SERIALIZER_IGBINARY) {
+            return igbinary_serialize($valze);
```
idr0id: should be `$value` instead of `$valze`

lib/Doctrine/Common/Cache/RedisCache.php
```diff
@@ -157,6 +180,24 @@ protected function doGetStats()
...
+        }
+
+        return serialize($value);
+    }
+
+    protected function unserialize($value)
+    {
+        if ($this->serializer === self::SERIALIZER_IGBINARY) {
+            return igbinary_unserialize($valze);
```
idr0id: should be `$value` instead of `$valze`

steffenweber:
Oops, both typos are fixed!

lib/Doctrine/Common/Cache/RedisCache.php
```diff
@@ -39,12 +45,28 @@ public function getRedis()
         return $this->redis;
     }
 
+    /**
+     * Gets the used serializer
+     *
+     * @return int
+     */
+    public function getSerializer()
```
lcobucci: Please declare the return type (instead of using doc block)

lib/Doctrine/Common/Cache/RedisCache.php
```diff
@@ -15,18 +15,24 @@
  */
 class RedisCache extends CacheProvider
 {
+    public const SERIALIZER_PHP = 1;
+    public const SERIALIZER_IGBINARY = 2;
+
     /** @var Redis|null */
     private $redis;
 
+    /** @var int */
```
lcobucci: This should be documented as nullable since the value is only configured when `setRedis()` gets called.

lib/Doctrine/Common/Cache/RedisCache.php
```diff
@@ -89,7 +111,8 @@ protected function doSaveMultiple(array $keysAndValues, $lifetime = 0)
         }
 
         // No lifetime, use MSET
-        return (bool) $this->redis->mset($keysAndValues);
+        $serializedKeysAndValues = array_map([$this, 'serialize'], $keysAndValues);
```
lcobucci: This variable isn't necessary

lcobucci (changes requested):
@steffenweber Is there any way to test that the race condition has actually been fixed?

lib/Doctrine/Common/Cache/RedisCache.php
steffenweber: I'll fix this.

lib/Doctrine/Common/Cache/RedisCache.php
steffenweber: You mean $serializedKeysAndValues? It's used on the next line.

lib/Doctrine/Common/Cache/RedisCache.php
lcobucci: Also, any reason for it be public?

steffenweber: Do you mean `public function getSerializer(): ?int` (which only seems to be supported since PHP 7.1)?

Ocramius: Yeah, it should unlikely be `public` on the cache instance

steffenweber:
@lcobucci Yes, you should be able to test that the race condition is fixed using the scripts I posted here: https://github.com/doctrine/cache/issues/258#issuecomment-389848729

lib/Doctrine/Common/Cache/RedisCache.php
steffenweber: Hmm, it's `public` to make `RedisCacheTest::testSerializerOptionWithOutIgbinaryExtension` work.

Ocramius: Can you change the test bit to use reflection, and then we look at how to make it work without reading this at all?

lcobucci: If that's ONLY for testing purposes we should simply not have the method at all... `self::assertAttributeSame()` helps us on that.

lib/Doctrine/Common/Cache/RedisCache.php
lcobucci: Yeah. I meant that, since the variable is only used once, we could simply inline the expression. Like:

```php
return (bool) $this->redis->mset(
    array_map([$this, 'serialize'], $keysAndValues)
);
```

lcobucci:
> @lcobucci Yes, you should be able to test that the race condition is fixed using the scripts I posted here: #258 (comment)

And could you try to add it as a test? Then we would be able to ensure that this would never happen again =).

steffenweber:
@lcobucci The test would somehow have to run the two scripts (or more generally speaking: functions) in parallel. According to some googling I've done, PHPUnit doesn't seem to support that. Unfortunately, I don't think this issue (or race conditions in general) can be unit-tested using PHPUnit.

Your other comments should be addressed by the latest update. :slightly_smiling_face:

lcobucci:
@steffenweber can you confirm that this branch is synced with `master`? That particular failure has already been fixed... If it's not you can just rebase your branch with the upstream and things should work. Let me know if you need help to handle that.

steffenweber:
@lcobucci Yes, I rebased my branch today (on top of commit 5e65fc393ecc17e395ac44d7b855dcf5e6864f34).

steffenweber:
I could try to trigger another test run (with `git commit --amend` and a force push) if you think that might help?

lcobucci:
@steffenweber thanks, I've triggered the build again... looks like an issue with @travis-ci.

lcobucci:
@steffenweber indeed: https://www.traviscistatus.com/incidents/pxj3py3xzn14

chrisdeeming:
@lcobucci do we have an ETA as to when this can be merged and released?

@steffenweber well done and thank you for working on this. If this is completed and fixes the problem fully we may have to include the changes from your fork in the next XF release.

rquadling:
For an 'attempt' to test a race condition, you could try using paratest. This essentially runs PHPunit in parallel. Make sure any unit tests are small classes. If you have a lot of large data providers in a single class, the whole class is run in 1 of the threads. Splitting that up into several classes means they are executed in parallel.

By having separate `@group` for parallel, you could build multiple identical test cases. 

Unfortunately, there is no guarantee that multiple runs of the suite would give the same result as you cannot control the exact execution order, so these would need to be 'soft' tests - may work - may fail. YMMV.

steffenweber:
@rquadling Thank you for the suggestion. I'll try using paratest in a couple of weeks (on holiday right now).

chrisdeeming:
Considering the complexities of testing this, and that the code and previous examples speak for themselves, my recommendation is that these changes are merged and if further issues arise in the future, then they can be addressed then.

FWIW we have been using a version of @steffenweber's fix (backported to 1.5.x) in production for a couple of weeks and even deployed it to our customers over a week ago. We have not had any continued reports of any issues, so I am therefore confident that the reported issue is now resolved.

steffenweber:
I've taken a look at paratest. It seems like the motivation behind paratest isn't testing race conditions but optimizing unit test performance.

Maybe it's somehow possible to make it work for testing race conditions as @rquadling described but I unfortunately wasn't able to figure out how exactly that could be achieved within a reasonable timeframe. 

Furthermore, I'm not sure whether Doctrine would be fine with adding a dependency on paratest.

If someone else can provide a test-case that would be awesome (but IMHO that isn't a "must have").

rquadling:
I wasn't sure if paratest could help. It is primarily about speedier unit testing, but as the tests are run in parallel, a suite of single test case unit tests (i.e. 1 test method in the unit test file) that ran in parallel that interact with the same resource could simulate the pattern as described in https://github.com/doctrine/cache/issues/258#issuecomment-389848729

chrisdeeming:
Does the test have to be a blocker, seeing as real world testing is proving that the proposed fix solves the evident issue?

Ocramius:
You might consider it narrow minded, but the maintenance work still lands
on our plate after a fix, so in order to prevent regressions it is
imperative to have a scenario covering this issue, sorry.

idr0id:
Looks like this PR is never been closed and the bug is never been fixed.

alcaeus:

PR #264, opened 2018-05-30
Add php-redis v4 support

Killium:
As desribed [here](https://github.com/phpredis/phpredis#exists),

Since php-redis v4 states that:
> This function took a single argument and returned TRUE or FALSE in phpredis versions < 4.0.0.

 We have to cast `doContains` functions with `bool` to return TRUE or FALSE.

### DIFF
--- a/lib/Doctrine/Common/Cache/RedisCache.php
+++ b/lib/Doctrine/Common/Cache/RedisCache.php
@@ -6,6 +6,7 @@ use Redis;
 use function array_combine;
 use function defined;
 use function extension_loaded;
+use function is_bool;
 
 /**
  * Redis cache provider.
@@ -96,7 +97,13 @@ class RedisCache extends CacheProvider
      */
     protected function doContains($id)
     {
-        return $this->redis->exists($id);
+        $exists = $this->redis->exists($id);
+
+        if (is_bool($exists)) {
+            return $exists;
+        }
+
+        return $exists > 0;
     }
 
     /**


jwage:
what does it return now?

Killium:
@jwage Now wether you return `1` or `true`, it will return `true`

cdaguerre:
@jwage SInce v4.0.0, it returns a `long`: the number of keys tested that do exist.

From the doc:
> Return value
long: The number of keys tested that do exist.

jwage:
Thanks. I think it would be better to make the condition explicit instead of casting. Something like:

```
return $this->redis->exists($id) > 0 ? true : false;
```

Killium:
You're right @jwage , it's done

cdaguerre:
Thats actually a bit weird because in versions of php-redis prior to v4 you check whether a bool is greater than 0.
```php
return true > 0 ? true : false;
```
Maybe something like this would be cleaner:
```php
$exists = $this->redis->exists($id);

if (is_bool($exists)) {
    return $exists;
}

return $exists > 0;

File: lib/Doctrine/Common/Cache/RedisCache.php [hunk 1]
```diff
@@ -96,7 +96,7 @@ protected function doSaveMultiple(array $keysAndValues, $lifetime = 0)
      */
     protected function doContains($id)
     {
-        return $this->redis->exists($id);
+        return $this->redis->exists($id) > 0 ? true : false;
```
Majkl578: `$this->redis->exists($id) > 0` would be enough.

Since it previously returned bool, `bool > integer` is invalid, this needs to be rewritten properly. I.e.:
```
$result = $this->redis->exists($id);

if (is_bool($result)) {
    return $result;
}

return $result > 0;
```

Before anyone says _it's a function call, it's slow!_ - [it won't be](https://github.com/php/php-src/blob/d0ee2a8254c69546c7191ac26cdc0499eba27acf/ext/standard/type.c#L232).

Tobion: You need to use `\is_bool` for opcache optimizations to take effect.

Majkl578: `use function is_bool`, but that is enforced by our CS so I didn't even mention it. :)

Killium:
@Majkl578 @Tobion @cdaguerre ,
I did update with @cdaguerre suggestion

lib/Doctrine/Common/Cache/RedisCache.php
Killium: @Tobion @Majkl578 Should I switch to `\is_bool` or keep the latest committed change ?

lib/Doctrine/Common/Cache/RedisCache.php
```diff
@@ -96,7 +96,13 @@ protected function doSaveMultiple(array $keysAndValues, $lifetime = 0)
      */
     protected function doContains($id)
     {
-        return $this->redis->exists($id);
+        $exists = $this->redis->exists($id);
+
+        if (\is_bool($exists)) {
```
jwage: This needs to be `use function is_bool;` at the top of the file. You can automatically handle this by running the following command:

```
./vendor/bin/phpcbf
```

Majkl578: Wouldn't use phpcbf due #265. :/

Killium: @Majkl578 So I just have to wait #265 to be merged ?

Majkl578: No, use `use function is_bool` and it's good to go.

jwage: Now that the use statement is there, you can remove the `\` before the ``is_bool`` call.

Killium:
Thank you @Majkl578 !

PR #270, opened 2018-06-19
ApcuCache driver name fixed

Vehsamrak:


Majkl578:
The failure is unrelated.

PR #283, opened 2018-10-13
CI: Test against PHP 7.3

Majkl578:


Majkl578:
FWIW couchbase extension still does not support PHP 7.3 in February 2019. :/

lcobucci:
We can simply skip the tests for that extension, it shouldn't prevent us from moving forward

Issue #292, opened 2018-12-03
PHP 7.2 compatibility


pavlepredic:
Given the [usage of a reserved word `object`](https://github.com/doctrine/cache/blob/master/lib/Doctrine/Common/Cache/RiakCache.php#L8) (reserved as of [PHP 7.2](https://secure.php.net/manual/en/reserved.other-reserved-words.php)), would it be fair to say that doctrine cache is incompatible with PHP 7.2 and this should therefore be reflected in the [composer.json](https://github.com/doctrine/cache/blob/master/composer.json#L27)?

lcobucci:
@pavlepredic that's an upstream issue IMO and riak cache update is being handled on https://github.com/doctrine/cache/pull/213. The PR was put on-hold by the contributor - it would be amazing if you could help things there :+1:.

I wouldn't say that `doctrine/cache` itself is incompatible, though.

pavlepredic:
Yes, I understand that the root cause is in Riak library, and I don't expect you to fix the underlying issue. I'm just suggesting that you specify in your library that PHP 7.2 is not supported. That would save people the headache of trying to figure out the weird errors stemming from some library they didn't even know they were using. In the long term, obviously it would be nice to either fix the issue upstream or remove RiakCache from doctrine/cache.

Ocramius:
Cache itself is 7.2 compatible, riak cache is not. We can't block the entire library due to one adapter, sorry.

pavlepredic:
There might just be a very simple solution for this problem. Simply alias the import here https://github.com/doctrine/cache/blob/master/lib/Doctrine/Common/Cache/RiakCache.php#L8 as `RickObject` and replace the occurrences of `Object` with `RickObject` in the code below.
I could make a PR if you want.

pavlepredic:
Here's the PR: https://github.com/doctrine/cache/pull/293. There's one more mention of `object` in the `isExpired` method typehint, but that's a lower-case object, so I'm guessing it's not supposed to be a Riak Object, right?

pavlepredic:
Actually it seems that it **is** supposed to be a Riak object, but it was wrongly corrected by a CS fixer: https://github.com/doctrine/cache/commit/a6af9992e910dc6434274e25cb0aa6b3c32d7314

Majkl578:
[...]

PR #293, opened 2018-12-03
Alias Riak\Object as RiakObject to prevent errors in PHP 7.2

pavlepredic:
As of PHP 7.2 `object` is a reserved word. Therefore, this line: https://github.com/doctrine/cache/blob/master/lib/Doctrine/Common/Cache/RiakCache.php#L8 triggers a PHP error in 7.2. This PR attempts to solve this issue by aliasing the import.

Issue: https://github.com/doctrine/cache/issues/292

### DIFF
--- a/lib/Doctrine/Common/Cache/RiakCache.php
+++ b/lib/Doctrine/Common/Cache/RiakCache.php
@@ -180,9 +180,9 @@ class RiakCache extends CacheProvider
     }
 
     /**
-     * Check if a given Riak Object have expired.
+     * Check if a given RiakObject has expired.
      */
-    private function isExpired(object $object) : bool
+    private function isExpired(RiakObject $object) : bool
     {
         $metadataMap = $object->getMetadataMap();
 
@@ -208,7 +208,7 @@ class RiakCache extends CacheProvider
      * @param string $vClock
      * @param array  $objectList
      *
-     * @return Object
+     * @return RiakObject
      */
     protected function resolveConflict($id, $vClock, array $objectList)
     {
--- a/lib/Doctrine/Common/Cache/RiakCache.php
+++ b/lib/Doctrine/Common/Cache/RiakCache.php
@@ -5,7 +5,7 @@ namespace Doctrine\Common\Cache;
 use Riak\Bucket;
 use Riak\Exception;
 use Riak\Input;
-use Riak\Object;
+use Riak\Object as RiakObject;
 use function count;
 use function serialize;
 use function time;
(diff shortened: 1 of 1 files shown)


Ocramius:
@pavlepredic can we force the autoloader to touch this file (as part of the test suite), so we can make sure that the parser doesn't choke on it?

lcobucci:
Wouldn't `php -l` be enough to validate that @Ocramius?

Ocramius:
@lcobucci yes :+1:

pavlepredic:
Are you waiting for anything from my side? Or can this be merged?

Ocramius:
I'm fine with this for now. Adding something like PHPStan (separate scope) would highlight it anyway, and it has to happen eventually.

Majkl578:
@Ocramius @lcobucci We should add [jakub-onderka/php-parallel-lint](https://github.com/JakubOnderka/PHP-Parallel-Lint) to our CI IMO.

Ocramius:
Meh, lib is tiny: additional installation time would be bigger than the gained speed.

Let's just run any other CI tool that causes the files to be parsed, and we're good.

PHPStan and/or Psalm being the obvious choice.

mktcode:
Hmmm for me it seems to work only if I rename the class itself. :(

Majkl578:
@mktcode This PR only fixes PHP 7.2+ compatibility (lint) on *our* side, Riak is indeed still broken - that is not in our hands, see https://github.com/basho/riak-php-client/issues/152.

fritzmg:
Judging by the comments in https://github.com/basho/riak-php-client/pull/153 & https://github.com/basho/riak-php-client/pull/154 `basho/riak-php-client` seems to be somewhat abandoned.

wuestkamp:
Nice, this seems to have fixed it for me PHP 7.2.13. Thx!

flexchar:
When could we expect a new release? 
It seems like this issue has been fixed but not yet live.

reedy:
> When could we expect a new release?
> It seems like this issue has been fixed but not yet live.

I'd agree, it'd be useful to have. Proposed backport of the two commits into 1.8.x in https://github.com/doctrine/cache/pull/319

reedy:
Soooo... This doesn't actually seem to fix the fatal - https://3v4l.org/JWocj

>Fatal error: Cannot use 'Riak\Object' as class name as it is reserved in /in/JWocj on line 185

> Judging by the comments in [basho/riak-php-client#153](https://github.com/basho/riak-php-client/pull/153) & [basho/riak-php-client#154](https://github.com/basho/riak-php-client/pull/154) `basho/riak-php-client` seems to be somewhat abandoned.

Sounds like a fork is needed if anyone wanted to actually work on it

pavlepredic:
Weird. I'm sure it fixed the issue back when I created the PR. Not sure anymore how I tested it, but all I cared about was Symfony console behaviour, which was throwing the fatal error until I applied this patch. But it's really difficult to reproduce after almost a year... Now indeed the PR seems pointless.

PR #306, opened 2019-05-22
Revert deprecation of memcache driver

bobvandevijver:
The memcache driver has been deprecated under the wrong assumption it is no longer supported on PHP 7.x in #267.

The memcache driver is actively maintained on https://github.com/websupport-sk/pecl-memcache and they are actually [working on releasing again on PECL](https://github.com/websupport-sk/pecl-memcache/issues/42). 

Currently, we're using this driver on php 7.3, and it works like a charm.


greg0ire:
It's not official and the lack of install instructions is a bit worrying, I wouldn't use it on a production system. Why exactly are you using it? Are there any feature you cannot find in the fully supported memcached extension, or is it just a way to avoid a migration?

bobvandevijver:
We have used the memcached extension in the past, which proved to be unstable and buggy on many occasions. The memcache is the one that never gave us any problem (apart from the installation as it is just a little bit more than apt install since we wanted to switch to PHP 7.3). 

The only reason it is not official seems to be due to the fact that the official maintainers never responded...

The [installation instructions](https://github.com/websupport-sk/pecl-memcache/issues/25) are pretty standard, just the default ones as with most php extensions.

greg0ire:
> The installation instructions are pretty standard, just the default ones as with most php extensions.

So you did all of these, including installing the dev version of php and git on your production server?

I do not know what the conditions are for us to support an extension, but I think having a package for most distributions should probably be a pre-requisite. I'm quite new to the team though, so I will let other maintainers chime in about this, they will have a better idea of what to do.

bobvandevijver:
Package is still available for the not yet released [Debian Buster](https://packages.debian.org/buster/php-memcache), which is actually based on the Github repository I linked earlier. We only built this ourselves as Debian Stretch does not supply PHP7.3 yet (comes with Buster), and we require at least PHP7.1 for the latest development.

Not yet tested (as we do not have it there yet), but on production we should be able to just place the build .so file without problems, and without placing the build tools.

greg0ire:
We discussed this internally and elected to merge this when that fork of memcache is available through PECL. That will be a good sign the project is getting healthy and supporting it is sensible for us.

bobvandevijver:
The latest version has just been released on [PECL](https://pecl.php.net/package/memcache/4.0.5). I've fixed the merge conflicts, so I guess this can be merged now :)

greg0ire:
@bobvandevijver the link in your messages produces a 404

milo526:
> @bobvandevijver the link in your message produces a 404

[4.0.5.1](https://pecl.php.net/package/memcache/4.0.5.1) has been released in the last few hours after which they seemingly removed the 4.0.5 version

https://pecl.php.net/package/memcache

greg0ire (approved):

SenseException (approved):

beberlei:
I merged this now

PR #307, opened 2019-05-26
Changed RedisCache to use pipeline in operations with multiple keys

yura3d:
Redis has a feature named _pipelining_ to group commands to execute. It reduces the number of requests (TCP or Unix socket) to Redis server by sending multiple commands per 1 request. More info on Redis site: https://redis.io/topics/pipelining

In this PR I've wrapped all the commands, that are sent to Redis in loops, into pipelines. This will have a positive effect on performance of methods that are working with multiple keys (_doFetchMultiple_ and _doSaveMultiple_).

### DIFF
--- a/lib/Doctrine/Common/Cache/RedisCache.php
+++ b/lib/Doctrine/Common/Cache/RedisCache.php
@@ -7,8 +7,8 @@ use function array_combine;
 use function array_diff_key;
 use function array_fill_keys;
 use function array_filter;
-use function array_intersect_key;
 use function array_keys;
+use function count;
 use function defined;
 use function extension_loaded;
 use function is_bool;
@@ -60,19 +60,18 @@ class RedisCache extends CacheProvider
         $fetchedItems = array_combine($keys, $this->redis->mget($keys));
 
         // Redis mget returns false for keys that do not exist. So we need to filter those out unless it's the real data.
-        $falseItemKeys = array_keys(array_filter($fetchedItems, static function ($item) {
+        $keysToFilter = array_keys(array_filter($fetchedItems, static function ($item) : bool {
             return $item === false;
         }));
-        if ($falseItemKeys) {
+
+        if ($keysToFilter) {
             $multi = $this->redis->multi(Redis::PIPELINE);
-            foreach ($falseItemKeys as $key) {
+            foreach ($keysToFilter as $key) {
                 $multi->exists($key);
             }
-            $nonExists        = array_filter($multi->exec(), static function ($exists) {
-                return ! $exists;
-            });
-            $nonExistItemKeys = array_intersect_key($falseItemKeys, $nonExists);
-            $fetchedItems     = array_diff_key($fetchedItems, array_fill_keys($nonExistItemKeys, true));
+            $existItems     = array_filter($multi->exec());
+            $missedItemKeys = array_diff_key($keysToFilter, $existItems);
+            $fetchedItems   = array_diff_key($fetchedItems, array_fill_keys($missedItemKeys, true));
         }
 
         return $fetchedItems;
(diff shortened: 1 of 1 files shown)


alcaeus:
Two questions:
* what versions of Redis is this available in? Do we want to check whether the feature is available?
* Can you add tests for this?

yura3d:
> what versions of Redis is this available in? Do we want to check whether the feature is available?

[MULTI/EXEC](https://redis.io/commands/multi) functionality is available since 1.2.0. We shouldn't add version check because the current version of RedisCache [relies](https://github.com/doctrine/cache/blob/17ec59c730888541191c63927791523be09fe3f7/lib/Doctrine/Common/Cache/RedisCache.php#L81) on [SETEX](https://redis.io/commands/setex) command, which was added in 2.0.0 and internally implemented through MULTI/EXEC calls. Also, all versions before 2.2.0 were released for about 10 years ago and now are [marked](https://code.google.com/archive/p/redis/downloads?page=4) as deprecated.

> Can you add tests for this?

The tests are already present:
https://github.com/doctrine/cache/blob/17ec59c730888541191c63927791523be09fe3f7/tests/Doctrine/Tests/Common/Cache/CacheTest.php#L78-L138
https://github.com/doctrine/cache/blob/17ec59c730888541191c63927791523be09fe3f7/tests/Doctrine/Tests/Common/Cache/RedisCacheTest.php#L15
Though I don't see the test for multiple saving with lifetime set. It is actual for all of cache drivers that supports multiple operations like MemcachedCache, not only for RedisCache. I can add it if it is needed.

lib/Doctrine/Common/Cache/RedisCache.php
```diff
@@ -55,17 +60,22 @@ protected function doFetchMultiple(array $keys)
         $fetchedItems = array_combine($keys, $this->redis->mget($keys));
 
         // Redis mget returns false for keys that do not exist. So we need to filter those out unless it's the real data.
-        $foundItems = [];
-
-        foreach ($fetchedItems as $key => $value) {
-            if ($value === false && ! $this->redis->exists($key)) {
-                continue;
+        $falseItemKeys = array_keys(array_filter($fetchedItems, static function ($item) {
```
Ocramius: ` : bool`

lib/Doctrine/Common/Cache/RedisCache.php
```diff
@@ -55,17 +60,22 @@ protected function doFetchMultiple(array $keys)
...
+            return $item === false;
+        }));
+        if ($falseItemKeys) {
+            $multi = $this->redis->multi(Redis::PIPELINE);
+            foreach ($falseItemKeys as $key) {
+                $multi->exists($key);
             }
-
-            $foundItems[$key] = $value;
+            $nonExists        = array_filter($multi->exec(), static function ($exists) {
```
Ocramius: `: bool`

Ocramius: `$exists` should have a type declaration

lib/Doctrine/Common/Cache/RedisCache.php
```diff
@@ -77,15 +87,14 @@ protected function doSaveMultiple(array $keysAndValues, $lifetime = 0)
...
-                if ($this->redis->setex($key, $lifetime, $value)) {
-                    continue;
-                }
-
-                $success = false;
+                $multi->setex($key, $lifetime, $value);
             }
 
-            return $success;
+            return ! array_filter($multi->exec(), static function ($setex) {
```
Ocramius: `: bool`

Ocramius: can `$setex` have a type declaration?

lib/Doctrine/Common/Cache/RedisCache.php
```diff
@@ -55,17 +60,22 @@ protected function doFetchMultiple(array $keys)
...
         // Redis mget returns false for keys that do not exist. So we need to filter those out unless it's the real data.
-        $foundItems = [];
-
-        foreach ($fetchedItems as $key => $value) {
-            if ($value === false && ! $this->redis->exists($key)) {
-                continue;
+        $falseItemKeys = array_keys(array_filter($fetchedItems, static function ($item) {
+            return $item === false;
+        }));
+        if ($falseItemKeys) {
```
Ocramius: Please put a new line before this block.

Ocramius:
Thanks @yura3d - going to release now

Issue #208, opened 2017-01-20
Memcached adapter should replace spaces


lcobucci:
As mentioned on https://github.com/doctrine/doctrine2/issues/4602 and  https://github.com/doctrine/doctrine2/pull/1424 Memcached has issues with keys that contains spaces and the ORM generates some of this keys for L2C.

However, as pointed by @Ocramius, the fix should be made on this package since only memcached has this issue.

alpha-and-omega:
This should be fixed. I wasted hours trying to understand why this fails:

```
<?php
require_once (__DIR__."/vendor/autoload.php");
$memcached = new Memcached();
$memcached->addServer('localhost', 11211);
$cache = new \Doctrine\Common\Cache\MemcachedCache();
$cache->setMemcached($memcached);
var_dump( $cache->fetch("some key") );
var_dump($cache->save('some key', 'my_data'));
var_dump( $cache->fetch("some key") );
```
bool(false)
bool(false)
bool(false)

gquemener:
Hello,

I ran into another issue related to memcached key limitation, so allow me to add my 2 cents (I could potentially fix both issues).

According to https://github.com/memcached/memcached/blob/master/doc/protocol.txt

> Keys
> ----
> 
> Data stored by memcached is identified with the help of a key. A key
> is a text string which should uniquely identify the data for clients
> that are interested in storing and retrieving it.  Currently the
> length limit of a key is set at 250 characters (of course, normally
> clients wouldn't need to use such long keys); the key must not include
> control characters or whitespace.

So there are a few steps required to sanitize the key before using it:
- Convert it into utf-8, if not already in this encoding
- Truncate it to 250 characters
- Strip whitespaces and control characters

One "simple" way of doing all that at once is to pass it to a hash function (as the FileCache adapter is doing).

Thus, I would suggest hashing (using fnv1a64 instead of sha256, because it is faster and offer a rather good non collision guarantee) the provided key before passing it to memcached.

This has the drawback of making manual memcached exploration harder, but fixes all the mentioned issues at once.

WDYT?

lcobucci:
Let's revive this!

@gquemener I think that drawback is quite relevant, especially because it's not the same use case as the `FileCache` adapter.

greg0ire:
Why not throw an exception instead?

gquemener:
> Let's revive this!
> 
> @gquemener I think that drawback is quite relevant, especially because it's not the same use case as the `FileCache` adapter.

How different is the use case? 

[...]

PR #308, opened 2019-05-31
Prevent invalid memcached key to be used

gquemener:
Here's the initial conversation: https://github.com/doctrine/cache/issues/208

This PR only ensures that memcached adapter provide valuable information while dealing with invalid key format.

TODO:

- [x] Fix CS
- [x] Generalize key format check (write operation only should be enough => saveMultiple)

### DIFF
--- a/lib/Doctrine/Common/Cache/MemcachedCache.php
+++ b/lib/Doctrine/Common/Cache/MemcachedCache.php
@@ -3,6 +3,10 @@
 namespace Doctrine\Common\Cache;
 
 use Memcached;
+use function array_keys;
+use function preg_match;
+use function strlen;
+use function strpos;
 use function time;
 
 /**
@@ -12,6 +16,8 @@ use function time;
  */
 class MemcachedCache extends CacheProvider
 {
+    public const CACHE_ID_MAX_LENGTH = 250;
+
     /** @var Memcached|null */
     private $memcached;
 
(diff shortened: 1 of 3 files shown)


tests/Doctrine/Tests/Common/Cache/MemcachedCacheTest.php
```diff
@@ -52,6 +53,25 @@ public function provideCacheIds() : array
...
+        $this->_getCacheDriver()->save($id, 1);
+    }
+
+    public function provideInvalidCacheIds() : array
+    {
+        return [
+            ['foo bar'], // contains space
+            ["\tfoo\n\r"], // contains control characters
+            [str_repeat('a', 251)], // long key
+        ];
```
greg0ire:
```
<?php
public function provideInvalidCacheIds() : array
{
    yield 'contains space' => ['foo bar'];
    yield 'contains control characters' => ["\tfoo\n\r"];
    yield 'long key' => [str_repeat('a', 251)];
```

lib/Doctrine/Common/Cache/MemcachedCache.php
```diff
@@ -78,6 +78,15 @@ protected function doContains($id)
...
     protected function doSave($id, $data, $lifeTime = 0)
     {
+        // https://github.com/memcached/memcached/blob/1.5.12/doc/protocol.txt#L41-L49
+        if (
+            strlen($id) > 250
+            || false !== strpos($id, ' ')
+            || !ctype_print($id)
+        ) {
+            throw new InvalidCacheId($id);
+        }
```
greg0ire: Maybe be even more helpful:
```
if (strlen($id) > 250) {
    throw InvalidCacheId::onKeyTooLong($id);
}
```

gquemener:
Hello,

Any special reasons blocking the merge of this PR, on which I could work on?

PR #311, opened 2019-07-01
apcu_fetch() emits E_NOTICE on read failure (undocumented):

Piskvor:
`Notice: apcu_fetch(): Error at offset 1938 of 7810 bytes`
In most modern code, this gets converted to an \ErrorException.
As this is a cache (and a miss is handled), discard the notice silently.


SenseException:
Thank you for your contribution. How exactly can this notice be triggered? It might be relevant if we can use a test to prevent regressions of this fix in the future.

alcaeus:

PR #332, opened 2019-11-23
PSR-6 implementation

derrabus:
The adapter added with this PR allows us to use Doctrine Cache as a PSR-6 implementation. This will allow us to decouple various Doctrine libraries from Doctrine cache.

#SymfonyHackday

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,7 +30,10 @@
         "mongodb/mongodb": "^1.1",
         "phpunit/phpunit": "^7.0",
         "predis/predis":   "~1.0",
-        "doctrine/coding-standard": "^6.0"
+        "doctrine/coding-standard": "^6.0",
+        "psr/cache": "^1.0",
+        "cache/integration-tests": "dev-master",
+        "symfony/phpunit-bridge": "^4.4 || ^5.0"
     },
     "suggest": {
         "alcaeus/mongo-php-adapter": "Required to use legacy MongoDB driver"
--- a/lib/Doctrine/Common/Cache/Psr6/CacheItem.php
+++ b/lib/Doctrine/Common/Cache/Psr6/CacheItem.php
@@ -62,7 +62,10 @@ final class CacheItem implements CacheItemInterface
     public function expiresAt($expiration): self
     {
         if (null !== $expiration && !$expiration instanceof \DateTimeInterface) {
-            throw new \TypeError(sprintf('Expected $expiration to be an instance of DateTimeInterface or null, got %s', \is_object($expiration) ? \get_class($expiration) : \gettype($expiration)));
+            throw new \TypeError(sprintf(
+                'Expected $expiration to be an instance of DateTimeInterface or null, got %s',
+                \is_object($expiration) ? \get_class($expiration) : \gettype($expiration)
+            ));
         }
 
         $this->expiration = $expiration;
--- a/lib/Doctrine/Common/Cache/Psr6/CacheItem.php
+++ b/lib/Doctrine/Common/Cache/Psr6/CacheItem.php
@@ -1,0 +1,94 @@
+<?php
+
+namespace Doctrine\Common\Cache\Psr6;
+
+use Psr\Cache\CacheItemInterface;
+
+final class CacheItem implements CacheItemInterface
+{
+    /** @var string */
+    private $key;
+    /** @var mixed */
+    private $value;
+    /** @var bool */
+    private $hit;
+    /** @var \DateTimeInterface|null */
+    private $expiration;
+
+    public function __construct(string $key, $data)
+    {
+        $this->key = $key;
+        $this->value = false === $data ? null : $data;
+        $this->hit = false !== $data;
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function getKey(): string
+    {
+        return $this->key;
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function get()
+    {
+        return $this->value;
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function isHit(): bool
+    {
+        return $this->hit;
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function set($value): self
+    {
+        $this->value = $value;
+
+        return $this;
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function expiresAt($expiration): self
+    {
+        if (null !== $expiration && !$expiration instanceof \DateTimeInterface) {
+            throw new \TypeError(sprintf('Expected $expiration to be an instance of DateTimeInterface or null, got %s', \is_object($expiration) ? \get_class($expiration) : \gettype($expiration)));
+        }
+
+        $this->expiration = $expiration;
+
+        return $this;
+    }
+    /**
+     * @inheritDoc
+     */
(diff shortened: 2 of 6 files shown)


composer.json
```diff
@@ -30,7 +30,10 @@
         "mongodb/mongodb": "^1.1",
         "phpunit/phpunit": "^7.0",
         "predis/predis":   "~1.0",
-        "doctrine/coding-standard": "^6.0"
+        "doctrine/coding-standard": "^6.0",
+        "psr/cache": "^1.0",
+        "cache/integration-tests": "dev-master",
+        "symfony/phpunit-bridge": "^4.4 || ^5.0"
```
greg0ire: Should we use this as an opportunity to use `"sort-packages": true`?

lib/Doctrine/Common/Cache/Psr6/CacheItem.php
```diff
@@ -0,0 +1,94 @@
...
+        return $this;
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function expiresAt($expiration): self
+    {
+        if (null !== $expiration && !$expiration instanceof \DateTimeInterface) {
+            throw new \TypeError(sprintf('Expected $expiration to be an instance of DateTimeInterface or null, got %s', \is_object($expiration) ? \get_class($expiration) : \gettype($expiration)));
```
greg0ire:
```suggestion
            throw new \TypeError(sprintf(
                'Expected $expiration to be an instance of DateTimeInterface or null, got %s',
                \is_object($expiration) ? \get_class($expiration) : \gettype($expiration)
            ));
```

lib/Doctrine/Common/Cache/Psr6/CacheItem.php
```diff
@@ -0,0 +1,94 @@
...
+    public function expiresAfter($time): self
+    {
+        if (null === $time) {
+            $this->expiration = null;
+        } elseif (is_numeric($time)) {
+            $this->expiration = new \DateTimeImmutable(sprintf('now +%d seconds', $time));
+        } elseif ($time instanceof \DateInterval) {
+            $this->expiration = (new \DateTimeImmutable())->add($time);
+        } else {
+            throw new \TypeError(sprintf('Expected $time to be either an integer, an instance of DateInterval or null, got %s', \is_object($time) ? \get_class($time) : \gettype($time)));
```
greg0ire:
```suggestion
            throw new \TypeError(sprintf(
                'Expected $time to be either an integer, an instance of DateInterval or null, got %s',
                \is_object($time) ? \get_class($time) : \gettype($time)
            ));
```

derrabus:
@Ocramius Would you prefer PSR-16 over PSR-6? In that case, we could use your library instead.

Ocramius:
PSR-6 was generally disapproved by doctrine during voting of the PSR itself, to the point that we left the organisation (among other reasons)

derrabus:
All right, that piece of information was missing at the Doctrine table here. So let's close this PR and move on with PSR-16.

greg0ire:
I'm currently trying out the roave library to get rid of the cache lib on doctrine/annotations.

nicolas-grekas:
I think PSR-6 is the way to go: having implemented both PSRs, I see only disadvantages to PSR-16. It provides no extensibility, and having iterable as type hints, it requires slower than needed code. If Symfony Cache is going to be a significant implementation target, the PSR-6 implementations are the most optimized, you'll figure out when benchmarking.

alcaeus:
@ocramius the point of this PR is not to have doctrine/cache add support for PSR-6 or 16 for the sake of supporting it: it is a stop-gap to replace it in our projects with a reliable cache implementation that is up the task in terms of support, performance, and extensibility. That’s why we agreed to replace it with PSR-6 for the time being.

alcaeus:
Reopening here: if we are to replace doctrine/cache with something else, we definitely need a PSR-6 adapter that allows using any PSR-6 cache in place of doctrine/cache. We may also want to add a PSR-16 adapter, but that's a story for a different PR.

We can then decide what to replace doctrine/cache with in each of the libraries that need a cache: different use-cases might call for different libraries.

Ocramius:
I'm still very skeptical on PSR-6 (would rather import PSR-16 - @asgrim, we should chat about this) since it's extremely bloated and over-complicated, but if it helps people moving away from our interfaces, it is still a decent addition.

Would rather see a PSR-16 implemented first, then eventually PSR-6 if really needed and if it cannot be avoided.

alcaeus:
I don't believe adding an adapter for one prohibits us from adding an adapter for the other: both cache interfaces have its merits depending on the use-case. As a generic cache-library on its way out, I don't think imposing such constraints is important: getting rid of a decade-old library that is no longer maintained, is.

Ocramius:
Yeah, I still don't want to maintain a PSR-6 implementation unless strictly absolutely required: mentioned multiple times that it is bloated and over-complicated.

PSR-16 first, then eventually PSR-6 if that's not sufficient.

derrabus:
> Would rather see a PSR-16 implemented first, then eventually PSR-6 if really needed and if it cannot be avoided.

Personally, I would pick the most complete interface, which is PSR-6. If we additionally wanted a PSR-16 implementation, constructing a wrapper would be trivial. If we directly implemented PSR-16, we cannot easily go back to PSR-6.

As an alternative to this PR, we could also use @Nyholm's Doctrine wrapper. That way, we would get PSR-6 and -16. https://github.com/php-cache/doctrine-adapter

nicolas-grekas:
Or https://github.com/symfony/cache/blob/master/DoctrineProvider.php
/ https://github.com/symfony/cache/blob/master/Adapter/DoctrineAdapter.php
btw.

lcobucci:
@derrabus there's no point in supporting both. Our current interface is quite similar to PSR-16, which makes it more feasible to rollout in our packages - specially if we plan to deprecate this library at certain point.

PSR-6 adds no value to the ORM nor to DBAL, only having to deal with a bloated API, which isn't quite good.

composer.json
lcobucci: We discussed about back then and prefer not to introduce phpunit-bridge. Testing for deprecations can be done with custom error handlers and we did in the ORM.

derrabus: > We discussed about back then and prefer not to introduce phpunit-bridge.

I agree. I currently need the bridge because the cache integration test that I‘ve used doesn‘t work without the bridge. My plan is to drop that dependency eventually.

asgrim: Aye, another +1 from me. I wasn't comfortable upgrading roave/doctrine-simple-cache to use this phpunit-bridge either, seems an unnecessary dependency, so assuming doctrine-simple-cache isn't assimilated by doctrine/cache, I may drop that dependency and just focus on our own tests. RE: https://github.com/php-cache/integration-tests/issues/105#issuecomment-545461444

derrabus:
I suggest that I pause my work here until the Doctrine project has come to a decision about which cache interface should be the replacement for this library. I have a personal preference, but I can do this work with either of them.

derrabus:
I'm going to close the PR now because I cannot work on such a topic without the unanimous support of Doctrine's core team.

derrabus:
The work on this topic has been resumed as #360.

Issue #327, opened 2019-11-12
CI doesn't have PHP 7.3 in the matrix


lcobucci:
We're simply missing it in `.travisci` :smile:

alcaeus:
I've added this to #325 as well since we'll be changing PHP version requirements anyways, but this can also be done separately.

alcaeus:
Fixed in #334.

PR #334, opened 2019-11-29
Support php 7.4

### LINKED ISSUE #325: Support PHP 7.4
With PHP 7.4 around the corner, we should ensure that the library works on PHP 7.4. The following tasks need to be handled:

* Drop support for PHP 7.1
* Require build for PHP 7.4 to pass
* Add PHP 7.3 to build matrix

alcaeus:
Fixes #325, #327.

This PR:
* adds PHP 7.3 to the build matrix
* requires build for PHP 7.4 to pass
* removes couchbase from tests since the extension causes segmentation faults.


tests/travis/Couchbase.sh
```diff
@@ -7,8 +7,8 @@ docker run -d --name couchbase-server -p 8091:8091 -p 11210:11210 webpt/couchbas
 
 pecl uninstall couchbase
 
-sudo wget -O/etc/apt/sources.list.d/couchbase.list http://packages.couchbase.com/ubuntu/couchbase-ubuntu1204.list
+sudo wget -O/etc/apt/sources.list.d/couchbase.list http://packages.couchbase.com/ubuntu/couchbase-ubuntu1404.list
```
stof: xenial is not 14-04, but 16-04

alcaeus: Changes were from me trying to get this to work again. I can drop these changes or remove the file entirely since we’re no longer calling it.

alcaeus: I've removed the file entirely.

alcaeus:
Good point. I've restored support for PHP 7.1 since we don't need to change anything there.

Issue #246, opened 2018-02-27
ChainCache save value with infinite lifetime when propagate to fastest cache


lchenay:
In case of item found in the chain of cache, a propagation of the value is made backward.
Those save are made without any TTL : https://github.com/doctrine/cache/blob/2b0e130afa0d70bafaf3da6c6d2022649562bc97/lib/Doctrine/Common/Cache/ChainCache.php#L50-L53
It lead to have infinite lifetime on those fastest cache ; and so many issue in production.

Here a test case to reproduce:
```php
    public function testFetchPropagateToFastestCacheWithTTL()
    {
        $cache1 = new ArrayCache();
        $cache2 = new ArrayCache();

        $cache2->save('bar', 'value', 1);

        $chainCache = new ChainCache([$cache1, $cache2]);

        self::assertFalse($cache1->contains('bar'));

        $result = $chainCache->fetch('bar');

        self::assertEquals('value', $result);
        self::assertTrue($cache1->contains('bar'));

        sleep(2);

        self::assertFalse($cache2->contains('bar'));
        self::assertFalse($cache1->contains('bar'));
        self::assertFalse($chainCache->contains('bar'));
    }
```

2 proposals that i see:
* Maybe we could have a short default ttl inside ChainCache that will be used in case of chain propagation?
* Get the remained ttl and propagate with this value ? (currently ttl is not exposed)

lchenay:
For proposal 2, many backend don't have a getter to retrieve the remain ttl.
A quick solution could be to wrap the cachedData in an object containing the {cachedData, ttl}.
But it will prevent to reuse the same Cache server without instancing a ChainCache as the stored data is not in the same format.

Without a new elegant solution on proposal 2, not sure it's worth it.

holtkamp:
mmm, just encountered this issue and it is actually quite "severe" right? It might also explain some crazy behaviour I have encountered for which I usually resorted to using a "TTL-specific" cacheEntryIdentifier...

**Description**
Mainly for my own understanding, a short description with an example.

Imagine an application running somewhere in the cloud, which [typically reboots once every 24 hours](https://devcenter.heroku.com/articles/dynos#restarting) with the following ChainCache:

- fast and volatile: is cleared on each server restart
- slow and persistant

Now suppose :
- a cache entry X is saved to the ChainCache with a TTL of 4 hours
  - cache entry X is saved to the fast cache with a TTL of 4 hours
  - cache entry X to the slow cache with a TTL of 4 hours
[...]

PR #335, opened 2019-12-11
Allow to indicate default TTL for downstream CacheProviders

holtkamp:
Allows users to resolve bug described in https://github.com/doctrine/cache/issues/246 where downstream CacheProviders are populated with cache entries with infinite lifetime, causing "strange" behavior: never expiring cache entries.

Behaviour is unchanged by default. 

Suggestions on how to test this are welcome 🤓

### DIFF
--- a/lib/Doctrine/Common/Cache/ChainCache.php
+++ b/lib/Doctrine/Common/Cache/ChainCache.php
@@ -15,6 +15,9 @@ class ChainCache extends CacheProvider
     /** @var CacheProvider[] */
     private $cacheProviders = [];
 
+    /** @var int */
+    private $defaultLifeTimeForDownstreamCacheProviders;
+
     /**
      * @param CacheProvider[] $cacheProviders
      */
@@ -23,6 +26,11 @@ class ChainCache extends CacheProvider
         $this->cacheProviders = $cacheProviders instanceof Traversable
             ? iterator_to_array($cacheProviders, false)
             : array_values($cacheProviders);
+    }
+
+    public function setDefaultLifeTimeForDownstreamCacheProviders(int $defaultLifeTimeForDownstreamCacheProviders) : void
+    {
+        $this->defaultLifeTimeForDownstreamCacheProviders = $defaultLifeTimeForDownstreamCacheProviders;
     }
 
     /**
@@ -48,7 +56,7 @@ class ChainCache extends CacheProvider
 
                 // We populate all the previous cache layers (that are assumed to be faster)
                 for ($subKey = $key - 1; $subKey >= 0; $subKey--) {
-                    $this->cacheProviders[$subKey]->doSave($id, $value);
+                    $this->cacheProviders[$subKey]->doSave($id, $value, $this->defaultLifeTimeForDownstreamCacheProviders);
                 }
 
                 return $value;
--- a/lib/Doctrine/Common/Cache/ChainCache.php
+++ b/lib/Doctrine/Common/Cache/ChainCache.php
@@ -16,7 +16,7 @@ class ChainCache extends CacheProvider
     private $cacheProviders = [];
 
     /** @var int */
-    private $defaultLifeTimeForDownstreamCacheProviders;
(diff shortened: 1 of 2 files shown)


lib/Doctrine/Common/Cache/ChainCache.php
```diff
@@ -15,6 +15,9 @@ class ChainCache extends CacheProvider
     /** @var CacheProvider[] */
     private $cacheProviders = [];
 
+    /** @var int */
+    private $defaultLifeTimeForDownstreamCacheProviders;
```
SenseException: This would override the default value `0` of `doSave()` with `null`. What if we keep the default value in here too?
```suggestion
    private $defaultLifeTimeForDownstreamCacheProviders = 0;
```

SenseException (changes requested):

lib/Doctrine/Common/Cache/ChainCache.php
holtkamp: Ah, I missed that default, totally true!

SenseException:
I forgot about the tests you asked for. Can you please add unit tests for now that covers the `$this->cacheProviders[$subKey]->doSave($id, $value, $this->defaultLifeTimeForDownstreamCacheProviders);` line for the default value and a set value?

holtkamp:
@SenseException 

>an you please add unit tests for now

Ah, yes, this is possible now. Will try to have a look in 2020 🤓

holtkamp:
@SenseException done!

Maybe the commits in this PR now needs to be squashed into one. I tried to do that numerous times and always failed, so pleas, be my guest to do this if required...

SenseException:
The case where no `setDefaultLifeTimeForDownstreamCacheProviders()` is called needs to be covered by tests.

Can you please use the current PHPUnit version and its method API for now? The approach didn't get feedback in #246 yet. We can decide afterwards about a PHPUnit update, which probably should be done in a separate PR.

holtkamp:
@SenseException ok, will try, but note that the "old" PHPUnit caused a deprecation error making the test fail... Hopefully I will find a way to circumvent / ignore that deprecation

tests/Doctrine/Tests/Common/Cache/ChainCacheTest.php
```diff
@@ -68,6 +68,51 @@ public function testFetchPropagateToFastestCache() : void
...
+        $chainCache = new ChainCache([$cache1, $cache2]);
+
+        $result = $chainCache->fetch('bar');
+
+        self::assertEquals('value', $result);
+    }
+
+    public function testFetchPropagatesToFastestCacheUsingIndicatedDefaultLifeTimeForDownstreamCacheProviders() : void
+    {
+        $pecificDefaultLifeTimeForDownstreamCacheProviders = 12345;
```
greg0ire: Did you really mean to use that `$` sign as an S? :sweat_smile:

holtkamp: Haha, oops, nope, will change!

holtkamp:
@SenseException, can we aim at wrapping this PR up? AFAIK it fixes a severe bug and the solution is quite trivial...

SenseException (approved):
Thank you for the last steps for this PR. The test method names are quite long, but for now this will work.

holtkamp:
Shameless ping to @SenseException and @greg0ire, is there anything merging this "reasonably" tested bug-fix?

greg0ire:
> Shameless ping to @SenseException and @greg0ire, is there anything merging this "reasonably" tested bug-fix?

blocking?

greg0ire:
Weird… the branch is aliased to 1.9, but there already is an 1.10 release… should I create a 1.11.0 milestone for this?

holtkamp:
@greg0ire it seems this merged PR has not landed in a release yet. Last release is https://github.com/doctrine/cache/releases/tag/1.10.0

Does this new PR justify a `1.10.1`? No functionality was added but it did fix a bug...

PR #340, opened 2020-03-11
Add support for atomic fetch with save if not found

cjunge-work:
APCu supports using `apcu_entry` to fetch an entry and save data using a generator if it's not found in the cache. These changes support that, as well as using (p)Redis pipelines with the atomic option to do the same.

I have implemented a fallback that is not atomic, but does the same for non-atomic providers.

In theory it would be possible to semi-replicate the atomic behaviour for providers that support transactions - while not strictly atomic, the implementation would be the similar with the fetch/save steps being grouped.

### DIFF
--- a/lib/Doctrine/Common/Cache/ApcuCache.php
+++ b/lib/Doctrine/Common/Cache/ApcuCache.php
@@ -2,14 +2,17 @@
 
 namespace Doctrine\Common\Cache;
 
+use RuntimeException;
 use function apcu_cache_info;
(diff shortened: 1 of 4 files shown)


cjunge-work:
For some strange reason I cannot get the coding standard build to work with the lambda function.

SenseException:
> For some strange reason I cannot get the coding standard build to work with the lambda function.

Can phpcbf fix this automatically?

alcaeus:

PR #341, opened 2020-05-18
Switch from Scrutinizer to Coveralls for coverage

greg0ire:
Took this as an occasion to update branch metadata, which I will merge up in `master`, where it has an effect.


.travis.yml
```diff
@@ -45,5 +45,6 @@ jobs:
       script:
         - ./vendor/bin/phpunit --exclude-group performance --coverage-clover clover.xml
       after_script:
-        - wget https://scrutinizer-ci.com/ocular.phar
-        - php ocular.phar code-coverage:upload --format=php-clover clover.xml
+        - wget https://github.com/php-coveralls/php-coveralls/releases/download/v2.2.0/php-coveralls.phar --output-document="${HOME}/bin/coveralls"
+        - chmod +x ${HOME}/bin/coveralls
+        - travis_retry coveralls -v --json_path /tmp/coveralls-upload.json --coverage_clover=clover.xml
```
greg0ire, line 50: @morozov I thought it would be more clever to put everything in the same step this time…
- easier to understand
- coveralls is not required to run the tests so shouldn't really be under `install`
- also, I noticed that I forgot to add `.coveralls` in `.gitattributes` on the DBAL, and thought that it might just be best not to have have a config file at all

I will open a PR on the DBAL if the above is ok with you.

greg0ire:
> Editing .doctrine-project.json and composer.json should be in its own PR and doesn't seem to be connected with the coveralls migration.

Why? It's in a separate commit already

SenseException:
Because I would have merged those to Coveralls unrelated changes by now. The changes for the docs are currently maintained in master branch.

greg0ire:
Anyway, this is on hold for now, see https://github.com/orgs/doctrine/teams/doctrinecore/discussions/15?from_comment=10

greg0ire:
We are going to use CodeCov.

PR #347, opened 2020-07-07
Patch default ttl for downstream cache providers

malarzm:
This is #335 again but targeting the `1.10.x` branch so the fix can be released as 1.10.2

cc @holtkamp


holtkamp:
@malarzm just for my understanding, why is this kind of "re-applying" required? It was already merged in master right? Could that not be released as 1.10.2 then? In case there is documentation about this subject in GIT always welcome 👍

malarzm:
I couldn't tag `1.10.2` on `master` because then your patch wouldn't be available on the `1.10.x` we use for maintenance. Any future bugfix will be targeting that branch and would miss your changes. Also we have a release bot (@doctrinebot) that has very strict expectations what should be where. Once I closed the 1.10.2 milestone the bot automatically tagged 1.10.2 on the `1.10.x` branch and did a merge up with changes to the next branch (master, #348).

PR #351, opened 2020-09-28
Update Composer, Travis Config and ChainCachetest for 8.0 support.

beberlei:
- Run travis update --ignore-platform-reqs to avoid mongodb problem
- Update Composer to allow PHPUnit 9
- Fix ChainCacheTest to not use deprecated PHPUnit API


alcaeus:
@beberlei is this still relevant for 1.11.x?

PR #353, opened 2020-10-17
SQLite3Cache: Statement can be false.

janbarasek:
In the event of an error in the set rights during the life of the application, the structure or content of the cached file may be damaged.

Doctrine does not currently throw an exception, but the entire request will end in a fatal error that simply cannot be processed further.

I suggest throwing an exception in such a case.

This type of error has happened 1000 times in my production application in the last month.

Real example:

Snímek obrazovky 2020-10-17 v 23 10 12

Thanks.

### DIFF
--- a/lib/Doctrine/Common/Cache/SQLite3Cache.php
+++ b/lib/Doctrine/Common/Cache/SQLite3Cache.php
@@ -2,12 +2,14 @@
 
 namespace Doctrine\Common\Cache;
 
+use RuntimeException;
 use SQLite3;
 use SQLite3Result;
 use const SQLITE3_ASSOC;
 use const SQLITE3_BLOB;
 use const SQLITE3_TEXT;
 use function array_search;
+use function error_get_last;
 use function implode;
 use function serialize;
 use function sprintf;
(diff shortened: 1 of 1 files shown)


lib/Doctrine/Common/Cache/SQLite3Cache.php
```diff
@@ -165,6 +165,10 @@ private function findById($id, bool $includeData = true) : ?array
             $idField
         ));
 
+        if ($statement === false) {
+            throw new \RuntimeException('SQLite internal database structure is broken: ' . (error_get_last()['message'] ?? ''));
```
greg0ire: > SQLite internal database structure is broken

You don't know that for sure, there might be other reasons for `prepare` returning `false`, right?

janbarasek: I think there may be more reasons and I don't know exactly how to find them. This will most likely be caused by a broken file.

The ʻerror_get_last` function should return information about the last error that occurred, for example, when trying to read a non-existent file to which there were no rights.

greg0ire: > I think there may be more reasons and I don't know exactly how to find them. 

Then maybe don't be so specific, or leave room for other explanations in the message. Also, it seems weird that you get `false` and not an exception. Are you using `doctrine/dbal`? It should configure PDO to throw on error: https://github.com/doctrine/dbal/search?q=ATTR_ERRMODE

PR #357, opened 2020-12-12
Maintenance work

greg0ire:


tests/Doctrine/Tests/Common/Cache/RedisCacheTest.php
```diff
@@ -27,36 +29,43 @@ protected function setUp() : void
...
-    protected function _getCacheDriver() : CacheProvider
+    public function testDeleteAll(): void
+    {
+        $this->markTestSkipped('this is probably a bug that needs to be fixed');
+    }
+
+    public function testDeleteAllAndNamespaceVersioningBetweenCaches(): void
+    {
+        $this->markTestSkipped('this is probably a bug that needs to be fixed');
+    }
```
greg0ire, line 66: Looks like I have found a bug with Redis

.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,104 @@
...
+    runs-on: "ubuntu-20.04"
+
+    strategy:
+      matrix:
+        php-version:
+          - "7.1"
+          - "7.2"
+          - "7.3"
+          - "7.4"
+          # - "8.0"
```
greg0ire, line 29: @alcaeus can you help with this? It fails because `alcaeus/mongo-php-adapter` cannot be installed, what should be done here?

alcaeus: That's odd, because 1.2.0 adds compatibility with PHO 8.0 🤔

greg0ire: Let me enable it again so that we can see the failure.

greg0ire: Ok I read to fast, looks like I just need to install the `mongo` extension

alcaeus: You need to install the `mongodb` extension - the legacy driver (a.k.a. the `mongo` extension) was never ported for PHP 7, which is why the adapter package exists.

greg0ire: I think this is what I did although I said otherwise :) I copied what you did in the ODM repo IIRC

SenseException (approved):

beberlei (approved):
Alright :)

alcaeus (approved):

PR #366, opened 2021-04-13
Import DoctrineProvider class to remove dependencies on symfony/cache

alcaeus:
This PR adds the DoctrineProvider class from symfony/cache. The class is marked as `@internal` and is designed for use in various other Doctrine packages that leak their caches (e.g. persistence). Users of doctrine/cache should not use this class at all.

The symfony/cache dev dependency is there to allow testing against an actual PSR-6 class - alternatively I could wrap a doctrine `ArrayCache` in a `CacheAdapter` and test against that, but it somehow seems wrong.

### DIFF
--- a/UPGRADE-1.11.md
+++ b/UPGRADE-1.11.md
@@ -11,6 +11,5 @@ There are two new classes to use in the `Doctrine\Common\Cache\Psr6` namespace:
   Doctrine cache implementations and switch to PSR-6.
 * The `DoctrineProvider` class allows using any PSR-6 cache as Doctrine cache.
   This implementation is designed for libraries that leak the cache and want to
-  switch to allowing PSR-6 implementations. This class is marked as internal and
-  should only be used during the transition phase of sunsetting doctrine/cache
-  support.
+  switch to allowing PSR-6 implementations. This class is design to be used
+  during the transition phase of sunsetting doctrine/cache support.
--- a/UPGRADE-1.11.md
+++ b/UPGRADE-1.11.md
@@ -4,3 +4,13 @@ doctrine/cache will no longer be maintained and all cache implementations have
 been marked as deprecated. These implementations will be removed in 2.0, which
 will only contain interfaces to provide a lightweight package for backward
 compatibility.
+
+There are two new classes to use in the `Doctrine\Common\Cache\Psr6` namespace:
+* The `CacheAdapter` class allows using any Doctrine Cache as PSR-6 cache. This
+  is useful to provide a forward compatibility layer in libraries that accept
+  Doctrine cache implementations and switch to PSR-6.
+* The `DoctrineProvider` class allows using any PSR-6 cache as Doctrine cache.
+  This implementation is designed for libraries that leak the cache and want to
+  switch to allowing PSR-6 implementations. This class is marked as internal and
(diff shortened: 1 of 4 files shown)


UPGRADE-1.11.md
```diff
@@ -4,3 +4,13 @@ doctrine/cache will no longer be maintained and all cache implementations have
...
 will only contain interfaces to provide a lightweight package for backward
 compatibility.
+
+There are two new classes to use in the `Doctrine\Common\Cache\Psr6` namespace:
+* The `CacheAdapter` class allows using any Doctrine Cache as PSR-6 cache. This
+  is useful to provide a forward compatibility layer in libraries that accept
+  Doctrine cache implementations and switch to PSR-6.
+* The `DoctrineProvider` class allows using any PSR-6 cache as Doctrine cache.
+  This implementation is designed for libraries that leak the cache and want to
+  switch to allowing PSR-6 implementations. This class is marked as internal and
```
nicolas-grekas: Isn't "internal" in conflict with the previous line, stating that this is for libraries?
What about `@deprecated` instead?

tests/Doctrine/Tests/Common/Cache/Psr6/DoctrineProviderTest.php
```diff
@@ -0,0 +1,45 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Doctrine\Tests\Common\Cache\Psr6;
+
+use Doctrine\Common\Cache\CacheProvider;
+use Doctrine\Common\Cache\Psr6\DoctrineProvider;
+use PHPUnit\Framework\TestCase;
+use Symfony\Component\Cache\Adapter\ArrayAdapter;
```
nicolas-grekas, line 17: you could use `CacheAdapter` instead here and get rid of `symfony/cache` in `require-dev`, isn't it?

alcaeus: > The symfony/cache dev dependency is there to allow testing against an actual PSR-6 class - alternatively I could wrap a doctrine ArrayCache in a CacheAdapter and test against that, but it somehow seems wrong.

;)

UPGRADE-1.11.md
alcaeus: It's mainly for internal libraries - I'm not sure how many non-doctrine libraries support injecting doctrine/cache implementations and leak those to the outside world. I wanted to add some kind of designation to avoid people thinking that this is code that will be maintained or should be used outside of compatibility layers, but I can also drop this to avoid triggering deprecation notices or anything else (e.g. when using the Symfony Debug class loader).

nicolas-grekas: the debug class loader will complain only if ppl extend the class, not if they just use it
but I think removing the annotation is fine also. The full component is sunsetted anyway, so the message should be sent not only for this class but for all of them in the lib.

alcaeus: I removed the designation as technically doctrine/persistence would be violating it already. If people want to shoot themselves in the foot with this, I won't complain.

nicolas-grekas (approved):

PR #368, opened 2021-04-27
Fix tests for 2.0 branch

alcaeus:
Re-introduces the abstract test class I inadvertently dropped and adds the ArrayCache to the test namespace so it's available to test the PSR-6 compatibility layer.


tests/Doctrine/Tests/Common/Cache/ArrayCacheTest.php
```diff
@@ -0,0 +1,52 @@
+<?php
+
+namespace Doctrine\Tests\Common\Cache;
+
+use Doctrine\Common\Cache\Cache;
+use Doctrine\Common\Cache\CacheProvider;
+
+class ArrayCacheTest extends CacheTest
```
derrabus, line 8: Testing a test fixture. That's thorough!

derrabus (approved):

PR #371, opened 2021-05-18
CacheAdapter: fix lifetime type

PavelJurasek:
Lifetime is computed as a difference of two floats (microtime) and then is passed to method annotated as integer. Lifetime is correctly cast to int when multiple items are saved (see L189), but it is not cast when single item is saved.


stof:
this project uses the merge-up workflow (as most Doctrine projects now). So this should be submitted against the 1.11 branch rather than being backported there after the fact.

PavelJurasek:
Ok, rebased onto 1.11.x and target of the PR changed too.

alcaeus (approved):

alcaeus:
Thanks @PavelJurasek!

PR #375, opened 2021-05-24
Multiple items manipulation in PSR-6 DoctrineProvider

yura3d:
Hi!

PSR-6 [DoctrineProvider](https://github.com/doctrine/cache/blob/1.11.x/lib/Doctrine/Common/Cache/Psr6/DoctrineProvider.php) doesn't use methods provided by PSR-6 `CacheItemPoolInterface` for multiple items manipulation, such as `getItems()`, `saveDeferred()`/`commit()` and `deleteItems()`. It may impact performance since base `CacheProvider` implementation of [doFetchMultiple()](https://github.com/doctrine/cache/blob/9c53086695937c50c47936ed86d96150ffbcf60d/lib/Doctrine/Common/Cache/CacheProvider.php#L215:L216)/[doSaveMultiple()](https://github.com/doctrine/cache/blob/9c53086695937c50c47936ed86d96150ffbcf60d/lib/Doctrine/Common/Cache/CacheProvider.php#L258:L259)/[doDeleteMultiple()](https://github.com/doctrine/cache/blob/9c53086695937c50c47936ed86d96150ffbcf60d/lib/Doctrine/Common/Cache/CacheProvider.php#L292:L293) simply converts 1 multiple operation to N single operations.

### DIFF
--- a/lib/Doctrine/Common/Cache/Psr6/DoctrineProvider.php
+++ b/lib/Doctrine/Common/Cache/Psr6/DoctrineProvider.php
@@ -19,8 +19,10 @@ use Traversable;
 
 use function array_combine;
 use function array_filter;
+use function array_flip;
 use function array_keys;
 use function array_map;
+use function array_replace;
 use function iterator_to_array;
 use function rawurlencode;
 
@@ -129,7 +131,7 @@ final class DoctrineProvider extends CacheProvider
      */
     protected function doSaveMultiple(array $keysAndValues, $lifetime = 0): bool
     {
-        $keys = array_keys($keysAndValues);
+        $keys        = array_keys($keysAndValues);
         $encodedKeys = array_map('rawurlencode', $keys);
 
         $items = $this->pool->getItems($encodedKeys);
--- a/lib/Doctrine/Common/Cache/Psr6/DoctrineProvider.php
+++ b/lib/Doctrine/Common/Cache/Psr6/DoctrineProvider.php
@@ -12,17 +12,12 @@
 namespace Doctrine\Common\Cache\Psr6;
 
 use Doctrine\Common\Cache\CacheProvider;
-use Psr\Cache\CacheItemInterface;
 use Psr\Cache\CacheItemPoolInterface;
 use Symfony\Component\Cache\Adapter\DoctrineAdapter as SymfonyDoctrineAdapter;
 use Traversable;
 
-use function array_combine;
-use function array_filter;
-use function array_flip;
 use function array_keys;
 use function array_map;
-use function array_replace;
 use function iterator_to_array;
 use function rawurlencode;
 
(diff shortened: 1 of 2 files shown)


lib/Doctrine/Common/Cache/Psr6/DoctrineProvider.php
```diff
@@ -66,6 +73,26 @@ protected function doFetch($id)
...
+     * @inheritDoc
+     */
+    protected function doFetchMultiple(array $keys): array
+    {
+        $items = $this->pool->getItems(array_map('rawurlencode', $keys));
+        if ($items instanceof Traversable) {
+            $items = iterator_to_array($items);
+        }
+
+        $items = array_combine($keys, $items);
```
stof: `$pool->getItems()` returns items keyed by the cache key. But it makes no guarantee that keys are in the same order.

stof (changes requested):

lib/Doctrine/Common/Cache/Psr6/DoctrineProvider.php
yura3d: Fixed https://github.com/doctrine/cache/pull/375/commits/d11f9218cc83498e19e378119d57eb954d438974

lib/Doctrine/Common/Cache/Psr6/DoctrineProvider.php
```diff
@@ -66,6 +75,31 @@ protected function doFetch($id)
...
+    {
+        $encodedKeys = array_map('rawurlencode', $keys);
+
+        $items = $this->pool->getItems($encodedKeys);
+        if ($items instanceof Traversable) {
+            $items = iterator_to_array($items);
+        }
+
+        // Sorting items by keys
+        $items = array_replace(array_flip($encodedKeys), $items);
```
stof: This code is basically looping over keys 4 times to build the final results. I think a foreach loop working in imperative way will be faster:

```php
$result = [];

foreach ($keys as $key)  {
    // To be checked whether redoing the rawurlencode is slower or faster than using a
    // single `array_combine` to create a mapping of keys to encoded keys, or comparing
    // that to building both the list and the map with a loop instead of the array_map
    $item = $items[rawurlencode($key)];

    if (!$item->isHit()) {
        continue;
    }

    $results[$key] = $item->get();
}
```

yura3d: Changed in 4155612

lib/Doctrine/Common/Cache/Psr6/DoctrineProvider.php
```diff
@@ -66,6 +70,33 @@ protected function doFetch($id)
...
+        $items = $this->pool->getItems(array_map('rawurlencode', $keys));
+        if ($items instanceof Traversable) {
+            $items = iterator_to_array($items);
+        }
+
+        $result = [];
+        foreach ($keys as $key) {
+            // To be checked whether redoing the rawurlencode is slower or faster than using a
+            // single `array_combine` to create a mapping of keys to encoded keys, or comparing
+            // that to building both the list and the map with a loop instead of the array_map
```
alcaeus: I think we can drop this comment - if someone notices a performance complaint they'll identify the cause soon enough.
```suggestion
```

lib/Doctrine/Common/Cache/Psr6/DoctrineProvider.php
```diff
@@ -92,6 +123,30 @@ protected function doSave($id, $data, $lifeTime = 0)
...
+    {
+        $items = $this->pool->getItems(array_map('rawurlencode', array_keys($keysAndValues)));
+        if ($items instanceof Traversable) {
+            $items = iterator_to_array($items);
+        }
+
+        foreach ($keysAndValues as $key => $value) {
+            // To be checked whether redoing the rawurlencode is slower or faster than using a
+            // single `array_combine` to create a mapping of keys to encoded keys, or comparing
+            // that to building both the list and the map with a loop instead of the array_map
```
alcaeus:
```suggestion
```

nicolas-grekas:
Honestly, I'm not sure this PR is worth it: it's a BC layer, aka something that ppl should stop using asap.
The added logic is non trivial. Better save ourselves from merging it, don't you think?

greg0ire:
Closing as per Nicolas' remark. Thanks for caring though.

PR #376, opened 2021-05-25
Fix wrong return type for DoctrineProvider::wrap

alcaeus:
We can return any cache, not just `DoctrineProvider` instances.


stof (approved):

greg0ire:
Thanks @alcaeus !

PR #381, opened 2021-07-07
Fix the PSR-6 wrapping for doctrine caches with a namespace

### LINKED ISSUE #380: Cache namespace from Setup::createCacheConfiguration gets lost
This is a cross post from doctrine/orm/issues/8829, because I'm not entirely sure if the bug is in the ORM package or in the cache layer. 

In any case, it's a very annoying problem for me at least, since it basically prevents you from running two separate sites (e.g. the typical dev/prod instances) on the same machine, since the sites override each other's cache. Feel free to close either this ticket or the orm one, because as I said, I'm not totally sure which repo is the source of the issue

stof:
If a namespace is set in a DoctrineProvider cache, we cannot unwrap the inner PSR-6 cache pool, as that would loose the namespacing feature. We need to add an additional wrapper instead.

Closes #380

### DIFF
--- a/tests/Doctrine/Tests/Common/Cache/Psr6/CacheAdapterTest.php
+++ b/tests/Doctrine/Tests/Common/Cache/Psr6/CacheAdapterTest.php
@@ -127,4 +127,43 @@ final class CacheAdapterTest extends CachePoolTest
         $adapter->commit();
         $adapter->commit();
     }
+
+    public function testNamespacingFeatureIsPreservedWithDoctrineProvider(): void
+    {
+        $wrapped = new ArrayAdapter();
+
+        $cacheApp1 = DoctrineProvider::wrap($wrapped);
+        $cacheApp1->setNamespace('app 1');
+
+        $cacheApp2 = DoctrineProvider::wrap($wrapped);
+        $cacheApp2->setNamespace('app 2');
+
+        $psrCacheApp1 = CacheAdapter::wrap($cacheApp1);
+        $psrCacheApp2 = CacheAdapter::wrap($cacheApp2);
+
+        $item = $psrCacheApp1->getItem('some key')->set('some value');
+        $psrCacheApp1->save($item);
+        self::assertFalse($psrCacheApp2->getItem('some key')->isHit());
+    }
+
+    /**
+     * @requires function Symfony\Component\Cache\DoctrineProvider::__construct
+     */
+    public function testNamespacingFeatureIsPreservedWithSymfonyDoctrineProvider(): void
+    {
+        $wrapped = new ArrayAdapter();
+
+        $cacheApp1 = new SymfonyDoctrineProvider($wrapped);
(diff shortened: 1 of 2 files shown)


stof:
Note that the ORM setup class should still be updated to apply the namespace on the PSR-6 layer directly when using the DoctrineProvider, to allow migrating away from it.

And I think we should probably trigger a deprecation when setting a non-empty namespace in the DoctrineProvider, as getting rid of doctrine/cache implies moving the needed namespacing to the PSR-6 layer. What do you think @alcaeus ?

greg0ire (dismissed):

greg0ire:
I think Andreas is busy. @derrabus , maybe you could give this an additional review? It could be nice to release this bugfix with the other one I merged today.

derrabus:
The fix _looks_ correct, but I think I would want a test case for that.

tests/Doctrine/Tests/Common/Cache/Psr6/CacheAdapterTest.php
```diff
@@ -127,4 +128,40 @@ public function testItemsAreFlushedToTheUnderlyingCacheOnce(): void
...
+        $psrCacheApp1->save($item);
+        self::assertFalse($psrCacheApp2->getItem('some key')->isHit());
+    }
+
+    /**
+     * @psalm-return iterable<string, array{Closure(CacheItemPoolInterface): (Cache|SymfonyDoctrineProvider)}>
+     */
+    public function doctrineProviderFactoryProvider(): iterable
+    {
+        yield 'Doctrine' => [Closure::fromCallable([DoctrineProvider::class, 'wrap'])];
```
greg0ire: The only flaw I see here is that we don't know for sure that this closure is going to return a `DoctrineProvider`… and there does not seem to be a way to get one for sure: https://github.com/doctrine/cache/blob/cebb0d13f607775ff4cca587995558dce87a026c/lib/Doctrine/Common/Cache/Psr6/DoctrineProvider.php#L31-L47

greg0ire: I have ditched the data provider I was using because the SymfonyDoctrineProvider no longer exists when using PHP 8, but the issue is still here: `DoctrineProvider::wrap` may not return a `DoctrineProvider`

greg0ire: I just realised we don't have static analysis setup on this project and probably never will, so this shouldn't be an issue.

derrabus (approved):

tests/Doctrine/Tests/Common/Cache/Psr6/CacheAdapterTest.php
stof: and that's fine. `DoctrineProvider::wrap` is *not* expected to return a DoctrineProvider but a Cache.

PR #382, opened 2021-07-13
Allow psr/cache v3

nicolas-grekas:
Although this lib is deprecated, it's going to fade away quite slowly I suppose.
Having it conflict with `"psr/cache": ">=3"` is going to block the adoption of the newer versions of the PSR.

This PR removes this conflict by adding a new `TypedCacheItem` class, since `CacheItem` is the only one that cannot be made compatible with psr/cache v3 without a BC break.

This PR also adds support for symfony/cache v6.

### DIFF
--- a/lib/Doctrine/Common/Cache/Psr6/TypedCacheItem.php
+++ b/lib/Doctrine/Common/Cache/Psr6/TypedCacheItem.php
@@ -1,0 +1,111 @@
+<?php
+
+namespace Doctrine\Common\Cache\Psr6;
+
+use DateInterval;
+use DateTime;
+use DateTimeInterface;
+use Psr\Cache\CacheItemInterface;
+use TypeError;
+
+use function get_class;
+use function gettype;
+use function is_int;
+use function is_object;
+use function microtime;
+use function sprintf;
+
+final class TypedCacheItem implements CacheItemInterface
+{
+    /** @var string */
+    private $key;
+    /** @var mixed */
+    private $value;
+    /** @var bool */
+    private $isHit;
+    /** @var float|null */
+    private $expiry;
+
+    /**
+     * @internal
+     */
+    public function __construct(string $key, mixed $data, bool $isHit)
+    {
+        $this->key   = $key;
+        $this->value = $data;
+        $this->isHit = $isHit;
+    }
+
+    public function getKey(): string
+    {
+        return $this->key;
+    }
+
+    public function get(): mixed
+    {
+        return $this->value;
+    }
+
+    public function isHit(): bool
+    {
+        return $this->isHit;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function set($value): static
+    {
+        $this->value = $value;
+
+        return $this;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function expiresAt($expiration): static
+    {
+        if ($expiration === null) {
+            $this->expiry = null;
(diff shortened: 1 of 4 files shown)


alcaeus:
@derrabus can you take a look? You removed this in #362, so you're probably in a better position to judge whether this makes sense. Thanks!

lib/Doctrine/Common/Cache/Psr6/TypedCacheItem.php
```diff
@@ -0,0 +1,111 @@
...
+
+    /**
+     * @internal
+     */
+    public function __construct(string $key, mixed $data, bool $isHit)
+    {
+        $this->key   = $key;
+        $this->value = $data;
+        $this->isHit = $isHit;
+    }
```
derrabus, line 37: Since we must not load this class on PHP 7 anyway (because of the `mixed` type), we could use PHP 8 syntax here.

```suggestion
    private ?float $expiry = null;

    /**
     * @internal
     */
    public function __construct(
        private string $key,
        private mixed $value,
        private bool $isHit
    ) {
    }
```

lib/Doctrine/Common/Cache/Psr6/TypedCacheItem.php
```diff
@@ -0,0 +1,111 @@
...
+
+    public function isHit(): bool
+    {
+        return $this->isHit;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function set($value): static
```
derrabus, line 57:
```suggestion
    public function set(mixed $value): static
```

lib/Doctrine/Common/Cache/Psr6/TypedCacheItem.php
```diff
@@ -0,0 +1,111 @@
...
+    public function expiresAt($expiration): static
+    {
+        if ($expiration === null) {
+            $this->expiry = null;
+        } elseif ($expiration instanceof DateTimeInterface) {
+            $this->expiry = (float) $expiration->format('U.u');
+        } else {
+            throw new TypeError(sprintf(
+                'Expected $expiration to be an instance of DateTimeInterface or null, got %s',
+                is_object($expiration) ? get_class($expiration) : gettype($expiration)
```
derrabus, line 76:
```suggestion
                get_debug_type($expiration)
```

lib/Doctrine/Common/Cache/Psr6/TypedCacheItem.php
```diff
@@ -0,0 +1,111 @@
...
+        if ($time === null) {
+            $this->expiry = null;
+        } elseif ($time instanceof DateInterval) {
+            $this->expiry = microtime(true) + DateTime::createFromFormat('U', 0)->add($time)->format('U.u');
+        } elseif (is_int($time)) {
+            $this->expiry = $time + microtime(true);
+        } else {
+            throw new TypeError(sprintf(
+                'Expected $time to be either an integer, an instance of DateInterval or null, got %s',
+                is_object($time) ? get_class($time) : gettype($time)
```
derrabus, line 97:
```suggestion
                get_debug_type($time)
```

derrabus (changes requested):

derrabus:
I think we should backport this change to v1 as well. Because we're adding a new cache item class, I'd file this change as a feature and release it as 1.12.0 and 2.1.0.

derrabus:
I'm closing this PR because having a `v3` branch on the origin repository confuses Composer. 😅 

Taking over in #383

PR #383, opened 2021-07-14
Allow psr/cache v3

derrabus:
Taking over from #382

As mentioned on that PR, I propose to bump to 1.12/2.1 and release this change as a feature.

### DIFF
--- a/.github/workflows/coding-standards.yml
+++ b/.github/workflows/coding-standards.yml
@@ -17,7 +17,7 @@ jobs:
     strategy:
       matrix:
         php-version:
-          - "7.4"
+          - "8.0"
 
     steps:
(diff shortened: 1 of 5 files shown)


.github/workflows/coding-standards.yml
```diff
@@ -17,7 +17,7 @@ jobs:
     strategy:
       matrix:
         php-version:
-          - "7.4"
+          - "8.0"
```
alcaeus, line 20: This will require changing branch protection rules to allow merging PRs again.

alcaeus (approved):
LGTM, Thank you!

nicolas-grekas:
Thank you both for taking over!

PR #360, opened 2021-02-26
Add PSR-6 adapter

nicolas-grekas:
A reminiscence of #332

Should ease sunsetting the lib (see #354)

Co-Authored-By: Alexander M. Turek <me@derrabus.de>
Co-Authored-By: Grégoire Paris <postmaster@greg0ire.fr>

PR #388, opened 2021-07-16
Update branch metadata to reflect current state

greg0ire:
The 1.x series is not maintained anymore. Also, there should not be a
2.2.x.

### DIFF
--- a/.doctrine-project.json
+++ b/.doctrine-project.json
@@ -5,20 +5,38 @@
     "docsSlug": "doctrine-cache",
     "versions": [
         {
+            "name": "2.1",
+            "branchName": "2.1.x",
+            "slug": "2.1",
+            "current": true
+        },
+        {
+            "name": "2.0",
+            "branchName": "2.0.x",
+            "slug": "2.0",
+            "maintained": false
+        },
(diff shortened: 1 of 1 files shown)


.doctrine-project.json
```diff
@@ -4,21 +4,39 @@
...
+            "name": "2.0",
+            "branchName": "2.0.x",
+            "slug": "2.0",
+            "maintained": false
+        },
+        {
+            "name": "1.12",
+            "branchName": "1.12.x",
+            "slug": "1.12",
+            "maintained": false
```
derrabus: I'd still flag the 1.12.x branch as maintained (for everything related to PSR-6) and merge bugfixes like #387 to it. Doing so shouldn't be much overhead since the PSR-6 wrappers are identical on 1.12.x and 2.1.x.

greg0ire: Oh right, I got fooled by the "merged into master" in https://github.com/doctrine/cache/pull/360 . Makes sense to retarget.

SenseException (approved):

derrabus (approved):

PR #391, opened 2021-07-18
Add example of using a PSR-6 cache wrapped in Doctrine Cache interface.

beberlei:
I want to add this explanation to the documentation and UPGRADE notice to help users upgrade more easily.

### DIFF
--- a/UPGRADE-1.11.md
+++ b/UPGRADE-1.11.md
@@ -17,11 +17,11 @@ There are two new classes to use in the `Doctrine\Common\Cache\Psr6` namespace:
 A full example to setup a filesystem based PSR-6 cache with symfony/cache
 using the `DoctrineProvider` to convert back to Doctrine's `Cache` interface:
 
-.. code-block:: php
+```php
+use Doctrine\Common\Cache\Psr6\DoctrineProvider;
+use Symfony\Component\Cache\Adapter\FilesystemAdapter;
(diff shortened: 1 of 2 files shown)


derrabus (approved):
Shall we target 1.12.x?

SenseException:
Because 1.12 is still maintained, I'd also suggest that it should target 1.12.x.

gnumoksha:
Hello. This is an important change. Are you planning to merge it soon?

Also:
1 - Is the `FilesystemAdapter` prefered in the same way the `PhpFileCache` was? [reference](https://www.doctrine-project.org/projects/doctrine-orm/en/2.9/reference/improving-performance.html#metadata-and-query-caches)
2 - Are you planning to update [this](https://www.doctrine-project.org/projects/doctrine-orm/en/2.9/reference/improving-performance.html#metadata-and-query-caches) and [this](https://www.doctrine-project.org/projects/doctrine-orm/en/2.9/reference/caching.html#phpfilecache)?

alcaeus:
@gnumoksha I'll re-target 1.12 then merge, but haven't gotten to it yet.

As for the other questions, I believe you should raise them in a discussion in the ORM project (see https://github.com/doctrine/orm/discussions), as they relate to how ORM integrates caching based on this library, but not directly related to the library.

PR #396, opened 2021-10-21
Abandon package

### LINKED ISSUE #395: mark doctrine/cache as abandoned on packagist?
the packagist information for doctrine/cache still looks like all is fine: https://packagist.org/packages/doctrine/cache

i suggest to change the description to explain the status, and also mark the package as abandoned so that composer install spits out a warning that users should move on to something else.

greg0ire:
Closes #395

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,7 +1,7 @@
 {
     "name": "doctrine/cache",
     "type": "library",
-    "description": "PHP Doctrine Cache library is a popular cache implementation that supports many different drivers such as redis, memcache, apc, mongodb and others.",
+    "description": "PHP Doctrine Cache library is a popular cache implementation that supports many different drivers such as redis, memcache, apc, mongodb and others. Abandoned in favor of any implementation of psr/cache, which is a standard",
     "keywords": [
         "php",
         "cache",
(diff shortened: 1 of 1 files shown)


dbu (dismissed):

composer.json
```diff
@@ -47,5 +47,6 @@
     },
     "autoload-dev": {
         "psr-4": { "Doctrine\\Tests\\": "tests/Doctrine/Tests" }
-    }
+    },
+    "abandoned": "psr/cache"
```
nicolas-grekas: psr/cache-implementation?

derrabus:
The ORM can not yet fully operate without the Doctrine Cache interfaces. I think, we should solve that before abandoning the package.

stof:
Indeed. We don't want all ORM users to receive a warning from composer about using an abandoned package.

dbu:
uh oh, was not aware. yeah until current orm runs without this package, it would be a bad idea to do a thing that triggers warnings.

greg0ire:
I should have known. Converting to draft until the right moment comes, which is when `doctrine/orm` 2.0 is no longer supported.

codedmonkey:
@alcaeus Any reason you abandoned the library without merging this pull request? It would be helpful to push people into the right direction.

We still use ORM2 in our project, and a notice in this library's README that you should be upgrading to ORM3 would also be helpful.

For context: https://www.doctrine-project.org/2025/10/08/an-update-on-the-orm-2-end-of-life.html

PR #398, opened 2022-05-10
Add reset() to PSR-6 DoctrineProvider

alexander-schranz:
When using a symfony application which run on swoole or roadrunner the services need to be resetted between the runs. Same for messenger:consume command services are resetted and require this call.

We did stumble else over problems with the caches and needed to add the `kernel.reset` tag to the doctrine caches: https://github.com/sulu/skeleton/pull/141/files

The reset method was implemented here and also reseted the Namespace correctly:

https://github.com/symfony/cache/blob/4c6747cf7e56c6b8e3094dd24852bd3e364375b1/DoctrineProvider.php#L49-L55

This seems not to be added when moved the DoctrineProvider from Symfony/Cache to this repository: https://github.com/doctrine/cache/pull/366

Symfony itself added the reset in this pull request. https://github.com/symfony/symfony/pull/24226. Maybe @nicolas-grekas can also gave insights about the get set Namespace here.

### DIFF
--- a/lib/Doctrine/Common/Cache/Psr6/DoctrineProvider.php
+++ b/lib/Doctrine/Common/Cache/Psr6/DoctrineProvider.php
@@ -129,6 +129,7 @@ final class DoctrineProvider extends CacheProvider
         if ($this->pool instanceof ResetInterface) {
             $this->pool->reset();
         }
+
         $this->setNamespace($this->getNamespace());
     }
 }
--- a/lib/Doctrine/Common/Cache/Psr6/DoctrineProvider.php
+++ b/lib/Doctrine/Common/Cache/Psr6/DoctrineProvider.php
@@ -58,6 +58,15 @@ final class DoctrineProvider extends CacheProvider
         return $this->pool;
     }
 
+    public function reset(): void
+    {
+        if ($this->pool instanceof ResetInterface) {
+            $this->pool->reset();
+        }
+
+        $this->setNamespace($this->getNamespace());
+    }
+
     /**
      * {@inheritdoc}
      */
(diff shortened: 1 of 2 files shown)


lib/Doctrine/Common/Cache/Psr6/DoctrineProvider.php
```diff
@@ -122,4 +123,12 @@ protected function doGetStats()
     {
         return null;
     }
+
+    public function reset()
```
alexander-schranz: We do not require an interface for this as the `service_resetter` of symfony doesn't require anything here: https://github.com/symfony/symfony/blob/6.1/src/Symfony/Component/HttpKernel/DependencyInjection/ServicesResetter.php

alexander-schranz:
The failing tests seems to be unrelated to this pull request, as they fail in MemomizeCache tests.

derrabus:
Please rebase. The CI should be green now. Also, can you please add a test that covers your change?

alexander-schranz:
@derrabus Thx, rebased!

derrabus:
I'd like to merge this. Can you please add tests?

lib/Doctrine/Common/Cache/Psr6/DoctrineProvider.php
```diff
@@ -122,4 +123,13 @@ protected function doGetStats()
...
```
derrabus:
```suggestion
    public function reset(): void
```

Let's move this new public method above the protected ones.

alexander-schranz:
@derrabus Test cases added.

derrabus (approved):

derrabus:
Released as 1.13.0 and 2.2.0.

alexander-schranz:
@derrabus Thank you!

PR #399, opened 2022-05-11
Run CI tests every week

alexander-schranz:
If creating a pull request like [here](https://github.com/doctrine/cache/pull/398). It is always nice to know if the tests did run on the target branch. In a repository which does not have a lot of changes it is maybe nice to run atleast once a week the tests via a cron.


stof:
Given that this package is being sunset, I don't see the need for that

derrabus:
I agree with @stof. Let's close. But thank you for your proposal.

